PR #51, opened 2014-03-05
Added createonly flags to avoid errors on second run

Yuav:
Changed default behavior of the pp file to only partition a drive once, instead of on every run.


tphoney:
Thanks for doing this.
Could you please try moving this into the provider. This should also have an associated unit test, be rebased, and get travis-ci passing.

tphoney:
Thanks for the work you have put in to this pull request. However due to inactivity and the following:
failing unit tests / no documentation / needs rebased / no tests added
This pull request is being closed. If you want to re-submit this pr with the necessary changes, we will look at the new pull request.

PR #57, opened 2014-05-11
Add swap partition support

hameedullah:
It seems like the swap partitions were not supported. So I have tried to add the swap partition support using just the logical_volume.pp and the filesystem provider.

filesystem provider needed to be updated, because it actually creates the filesystem, so in case of swap we need to use mkswap utility instead of mkfs.*

In logical_volume.pp I have made sure that /etc/fstab has the workable and correct entry, plus the swap is added to system swap using swapon -a. The only issue here is in /etc/fstab that the mount point for swap should be none.


mhaskel:
@hameedullah thanks for the contribution. Looks like these commits have been pulled in to #89, so I'm going to close this pull request and discussion/merging will happen there.

PR #69, opened 2014-08-21
Resolve ambiguous reference warnings in Geppetto.

smithj4:
The lvm module contains custom providers and definitions with identical
names that caused "Ambiguous reference" warnings in Eclipse Geppetto.


tphoney:
Agreeing with @ghoneycutt closing. Thanks for making the PR though.

PR #71, opened 2014-09-10
Add install lvm2

rhoml:
If lvm2 package is not installed on the server the LVM won't be able to be configured.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -2,7 +2,19 @@
 #
 class lvm (
   $volume_groups = {},
+  $version       = 'installed'
 ) {
+
+  $real_provider = $::osfamily ? {
+    'Debian' => 'apt',
+    'RedHat' => 'yum'
+  }
+
+  package {
+    'lvm2':
+      ensure   => $version,
+      provider => $real_provider
+  }
 
   validate_hash($volume_groups)
 


mhaskel:
@rhoml thanks for this contribution, and apologies that it took us so long to review this. While I like optionally adding the package dependency, I'd prefer to see it implemented with one of our more standard patterns. Rather than adding two extra classes to install the package, you can just add the package resource in init.pp (if on a *deb system). Also, rather than checking for `$::operatingsystem`, you can check for `$::osfamily == 'Debian'`. Please also don't wrap the package resource in a `if ! defined()` block. 

Once you've fixed these changes, the commits should be rebased against master and squashed down to a single commit. Thanks again!

rhoml:
Will update the idea during this weekend.

mhaskel:
Great, thanks @rhoml !

rhoml:
@mhaskel Some tests are failing also in Master or did I do something wrong?

manifests/init.pp
```diff
@@ -2,8 +2,20 @@
 #
 class lvm (
   $volume_groups = {},
+  $version       = 'installed'
```
mhaskel: This variable should be named `$package_ensure`.

manifests/init.pp
```diff
@@ -2,8 +2,20 @@
...
 ) {
 
+  $real_provider = $::osfamily ? {
+    'Debian' => 'apt',
+    'RedHat' => 'yum'
+  }
```
mhaskel: This block should be removed, you shouldn't need to specify the provider, puppet defaults should be fine.

manifests/init.pp
```diff
@@ -2,8 +2,20 @@
...
+
+  package {
+    'lvm2':
```
mhaskel: This should be on the previous line.

manifests/init.pp
```diff
@@ -2,8 +2,20 @@
...
   $volume_groups = {},
+  $version       = 'installed'
 ) {
 
+  $real_provider = $::osfamily ? {
+    'Debian' => 'apt',
+    'RedHat' => 'yum'
+  }
+
+  package {
```
mhaskel: Initially wasn't this patch only to install the package on deb systems? If so, the package resource should be wrapped in an `if $::osfamily == 'Debian'` block.

tphoney:
Hi @rhoml , thanks for the work you have put in to this pull request. However due to inactivity and failing unit tests this pull request is being closed. If you want to re-submit this PR with the necessary changes, we will gladly look at the new pull request.

PR #78, opened 2014-10-18
Extents management fix and encryption added

apengue:
I fixed the extent setting usage from hiera, and also added cryptsetup to encrypt the logical volume using a keyfile.
Regarding security, it is not the optimal solution and it has to be improved but I'd like also to share the initial idea.

### DIFF
--- a/lib/puppet/provider/logical_volume/lvm.rb
+++ b/lib/puppet/provider/logical_volume/lvm.rb
@@ -8,20 +8,22 @@ Puppet::Type.type(:logical_volume).provide :lvm do
              :resize2fs  => 'resize2fs',
              :umount     => 'umount',
              :blkid      => 'blkid',
-             :dmsetup    => 'dmsetup'
+             :dmsetup    => 'dmsetup',
+             :cryptsetup => 'cryptsetup',
+             :echo       => 'echo',
+             :fsck	 => 'fsck'
 
     optional_commands :xfs_growfs => 'xfs_growfs',
                       :resize4fs  => 'resize4fs'
 
     def create
         args = ['-n', @resource[:name]]
-        if @resource[:size]
+        if @resource[:extents]
+            args.push('--extents', @resource[:extents])
+        elsif @resource[:size]
             args.push('--size', @resource[:size])
         elsif @resource[:initial_size]
             args.push('--size', @resource[:initial_size])
-        end
-        if @resource[:extents]
-            args.push('--extents', @resource[:extents])
         end
 
         if !@resource[:extents] and !@resource[:size] and !@resource[:initial_size]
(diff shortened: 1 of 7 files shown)


mhaskel:
@apengue thanks for this contribution, and apologies it took so long for us to review this. I'll have more comments soon (and inline), but as a general comment this seems like it would best be broken up to multiple PRs since the extents work doesn't relate to the encryption work.

lvm.yaml
```diff
@@ -0,0 +1,18 @@
...
+        encryptionkeyfile: '/root/keyfile'
+#        size: 5G
+        size: 4732M
+#        extents: 90%FREE
+        #extents: 100%FREE
+        mountpath: /data
+lbs::gmlc::mlp_ip: '192.168.1.126'
+lbs::gmlc::gmlc01_ip: '192.168.0.140'
+lbs::gmlc::gmlc02_ip: '192.168.0.41'
+lbs::gmlc::gmlc_sim_ip: '192.168.0.42'
```
mhaskel: This file should not be added.

manifests/init.pp
```diff
@@ -1,9 +1,6 @@
-# == Class: lvm
-#
-class lvm (
-  $volume_groups = {},
+class lvm(
+  $volume_groups    = {},
```
mhaskel: These changes should be reverted.

manifests/volume.pp
```diff
@@ -115,11 +116,12 @@
...
-        volume_group => $vg,
-        size         => $size,
-        extents      => $extents,
-        require      => Volume_group[$vg]
+        ensure        => present,
+        volume_group  => $vg,
+        size          => $size,
+        extents       => $extents,
+        encryptionkeyfile => $encryptionkeyfile,
+        require       => Volume_group[$vg]
```
mhaskel: `=>` alignment

lib/puppet/provider/logical_volume/lvm.rb
```diff
@@ -8,21 +8,23 @@
              :resize2fs  => 'resize2fs',
              :umount     => 'umount',
              :blkid      => 'blkid',
-             :dmsetup    => 'dmsetup'
+             :dmsetup    => 'dmsetup',
+             :cryptsetup => 'cryptsetup',
+             :echo       => 'echo',
+             :fsck	 => 'fsck'
```
mhaskel: `fsck` should not be added.

lib/puppet/provider/logical_volume/lvm.rb
```diff
@@ -108,13 +116,22 @@ def size=(new_size)
...
+               path="/dev/mapper/enc-#{@resource[:volume_group]}-#{@resource[:name]}"
+               cryptsetup("-q", "-d", "#{@resource[:encryptionkeyfile]}", "open", "/dev/#{@resource[:volume_group]}/#{@resource[:name]}","enc-#{@resource[:volume_group]}-#{@resource[:name]}")
+               cryptsetup("-q", "resize", encpath)
+            end
             blkid_type = blkid(path)
             if command(:resize4fs) and blkid_type =~ /\bTYPE=\"(ext4)\"/
               resize4fs( path) || fail( "Cannot resize file system to size #{new_size} because resize2fs failed." )
             elsif blkid_type =~ /\bTYPE=\"(ext[34])\"/
-              resize2fs( path) || fail( "Cannot resize file system to size #{new_size} because resize2fs failed." )
+              fsck("-y", "-f",path) && resize2fs( path) || fail( "Cannot resize file system to size #{new_size} because resize2fs failed." )
```
mhaskel: `fsck` should not be added

manifests/volume_group.pp
```diff
@@ -1,11 +1,8 @@
-# == Define: lvm::volume_group
-#
-define lvm::volume_group (
+define lvm::volume_group(
```
mhaskel: These changes should be reverted

mhaskel:
@apengue actually, on further review I don't think this belongs as part of the lvm module. It seems like the encryption work makes more sense to have in a separate module since it isn'y actually tied to lvm. Since the extents work has already been fixed in a different pull request, I'm going to close this PR. Thank you for the contribution though!

PR #81, opened 2014-11-16
escape dashes in lv name for dmsetup remove

jhofeditz:
We use dashes in our LV names and they are automatically escaped by LVM (The dashes are doubled). When we create VG myvg and LV my-lv, the lvcreate device is /dev/mapper/myvg-my--lv.

This only affects dmsetup remove. The other LVM commands and outputs use the unescaped names.
I'm not sure if a new set of tests should be created or if the current create and destroy test should be modified to use myvg/my-lv so I didn't make any spec changes.


mhaskel:
Great, thanks @jhofeditz !

PR #90, opened 2014-12-22
Support for readahead count when creating new volumes

jayc89:


wallies:
+1 for getting this merged, as I dont really want to be changing the repo and need this setting for mongo servers

mhaskel:
@jayc89 thanks for the contribution! In general this looks good, but I have one question which I'll put in an inline comment. If you can respond/fix that as relevant we should be good to merge this.

lib/puppet/type/logical_volume.rb
```diff
@@ -77,6 +77,15 @@
     end
   end
 
+  newparam(:readahead) do
+    desc "The readahead count to use for the new logical volume."
+    validate do |value|
+      unless value =~ /^([0-9]+|Auto|None)/i
```
mhaskel: Does lvcreate care about the casing for auto or none? If it does this parameter should have a munge block to fix the case.

seif: I just tried lvcreate with different casing of nOnE and AuTo, it doesn't care.

seif:
Would be great if this gets merged indeed! I just needed to set readahead and found this pull request!

mhaskel:
Great, thanks for the contribution @jayc89 !

seif:
@mhaskel sorry to pester, any ETA on when this would get pushed to the forge?

JamieCressey:
Hey @mhaskel, any idea when the next release will be cut? Thanks!

mhaskel:
Woops, sorry @seif @JamieCressey, I missed these pings. I'm not sure if we can get this out next week, but we should be able to do the following week at the latest. Thanks!

seif:
@mhaskel no worries, using git ref for now.

PR #93, opened 2015-01-28
Fix filesystem type detection

ckaenzig:
Sometimes `blkid` shows a `SEC_TYPE="ext2"` in addition to `TYPE="ext3"`.
It has something to do with a filesystem being backwards compatible with
ext2. This commit fixes the regex that matches TYPE=.. to prevent it
from matching SEC_TYPE.

Unfortunately I couldn't figure out how to add a test case for this. Any advice welcome if you think a test is necessary, thanks!

### DIFF
--- a/lib/puppet/provider/filesystem/lvm.rb
+++ b/lib/puppet/provider/filesystem/lvm.rb
@@ -16,7 +16,7 @@ Puppet::Type.type(:filesystem).provide :lvm do
     end
 
     def fstype
-        /TYPE=\"(\S+)\"/.match(blkid(@resource[:name]))[1]
+        /\bTYPE=\"(\S+)\"/.match(blkid(@resource[:name]))[1]
     rescue Puppet::ExecutionFailure
         nil
     end


lib/puppet/provider/filesystem/lvm.rb
```diff
@@ -16,7 +16,7 @@ def destroy
     end
 
     def fstype
-        /TYPE=\"(\S+)\"/.match(blkid(@resource[:name]))[1]
+        / TYPE=\"(\S+)\"/.match(blkid(@resource[:name]))[1]
```
raphink: Shouldn't that be 

```
 /\bTYPE=\"(\S+)\"/.match(blkid(@resource[:name]))[1]
```
?

saimonn: +1 for boundary matching, instead of space character.

mhaskel:
Thanks for the contribution @ckaenzig! I agree with @raphink's comment. Can you update that and then we should be good to merge this.

ckaenzig:
Thanks @mhaskel. I also agree and added acomit to change this. Should I squash both commits into one? I never know which is best.

ckaenzig:
Squashed on @raphink's advice.

mhaskel:
Thanks for the contribution @ckaenzig !

PR #55, opened 2014-04-23
Bug fixes for path and missing dependency errors

ITBlogger:
Fixes bugs related to explicit path missing from mkdir and test as well as dependency error from requiring File check that doesn't already exist in catalog.

PR #94, opened 2015-02-04
Allow different values for pass and dump

esalberg:
Allow root filesystems to have pass specified as 1 instead of 2; allow the default dump value to be overridden if desired.

### DIFF
--- a/manifests/logical_volume.pp
+++ b/manifests/logical_volume.pp
@@ -6,6 +6,8 @@ define lvm::logical_volume (
   $initial_size      = undef,
   $ensure            = present,
   $options           = 'defaults',
+  $pass              = '2',
+  $dump              = '1',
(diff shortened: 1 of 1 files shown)


esalberg:
Noted previously in https://github.com/puppetlabs/puppetlabs-lvm/pull/55 but spun off due to that mostly dealing with other changes.

I didn't add any extra documentation or spec tests - please let me know if there's anything you need.  I tried to keep this as simple as possible.  Functionally tested on RHEL 6.5 with Puppet 3.3.2 (using Hiera for volume definition).

mhaskel:
Fixes puppetlabs/puppetlabs-lvm#55

mhaskel:
@esalberg it doesn't look like existing tests/docs need any updates for this. thanks for the fix!

PR #95, opened 2015-02-06
Made logical_volume $size an optional parameter

campos-ddc:
Not setting it should use the entire physical drive.


esalberg:
I was just going to query if a change like this would be acceptable.  Since it's not possible to set "undef" in hiera, it would be great to have that in the code as default.

PR #98, opened 2015-02-19
Add timeout of 30 to facts

esalberg:
Adds a timeout to the lvm facts to handle more gracefully situations where lvm may not be behaving properly.

### DIFF
--- a/lib/facter/lvm_support.rb
+++ b/lib/facter/lvm_support.rb
@@ -1,6 +1,6 @@
 # lvm_support: true/nil
 #   Whether there is LVM support (based on the presence of the "vgs" command)
-Facter.add('lvm_support') do
+Facter.add('lvm_support', :timeout => 30) do
   confine :kernel => :linux
 
   setcode do
@@ -12,7 +12,7 @@ end
 # lvm_vgs: [0-9]+
 #   Number of VGs
 vg_list = []
-Facter.add('lvm_vgs') do
+Facter.add('lvm_vgs', :timeout => 30) do
   confine :lvm_support => true
   vgs = Facter::Util::Resolution.exec('vgs -o name --noheadings 2>/dev/null')
   if vgs.nil?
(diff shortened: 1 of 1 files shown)


esalberg:
We have had a couple of situations where maintenance work has caused disks to be removed inappropriately, causing "vgs" to hang.  When this happens, the lvm facts (I believe) both hang Facter and spawn multiple processes.

While this isn't Puppet's fault, it seems to make sense to put a timeout on the fact so that eventually it will give up / the process will end.  I chose 30 seconds because it seemed long enough as not to impact anything but short enough that unless there are a ton of volume groups, etc. it will probably end before the next Puppet run at least.  I have no problem with that value being changed, or if it's deemed that the timeout isn't needed on every Facter add in the facts file - just let me know and I'll update the pull request.

tphoney:
I think that timeout should be a variable that is set, this will allow it to be configurable.

tphoney:
Thanks for the work by the way.

esalberg:
I'm actually not sure how to set a variable from Puppet code in a pluginsync fact - I've just used other facts / top level variables.  Is this something simple that I'm missing?  The only thing I can think of right now would be to create a separate static fact which would have access to Puppet code variables, e.g.:

      file { 'facts.d/lvm_timeout':
        ensure  => 'file',
        path    => '/etc/puppetlabs/facter/facts.d/lvm_timeout.txt',
        content => "lvm_timeout=${timeout}",
        mode    => '0644',
      }

And then reference $::lvm_timeout in lvm_support.rb.

However, this seems kind of clunky, and it could take two puppet runs to get the timeouts set.

tphoney:
Adding a timeout of 30 to facts, will increase the run time significantly for each run. 
This also needs to be re-based.

bmjen:
Thanks for the contribution @esalberg. Having a configurable timeout isn't really much of an issue. However, this custom fact will no longer work with Facter 3.0. Are you able to refactor this code to be compatible with Facter 3.0 given this updated documentation. 
https://docs.puppetlabs.com/facter/3.0/custom_facts.html#execution-timeouts

hunner:
Closing due to inactivity. Feel free to reopen!

PR #100, opened 2015-03-10
Ordering of resources and removal of some defaults in class lvm

TorLdre:
Hi, thank you all for this fine module! I have a couple of issues though (and a proposal for solution).

1) When using the lvm class I discovered that the resources are created in random order, so what happens is, the code fails because the vg can't be created on non-existent pv, mkfs can't create file systems on none-existent logvols and so on. This happens and is reproducable when provision several devices. Another example, vgcreate will fail if the drive has an msdos label (in lack of --yes parameter), which happens often, I guess (given eg clearpart --all in anaconda). However, if those drives are pvcreated first, it works. These problems are addressed in manifests/volume_group.pp line 13 and manifests/logical_volume.pp line 58.

2) The other issue is with the creation of logical volumes using the lvm class. By the current design one has to provide size parameter (while extents should suffice), and it's assumed that the user want's a filesystem and mountpoint, and the device mounted afterwards. However, there are several use cases where this is not what the user want's (me). What I want is a logical volume without file system, and therefore no mountpoints or mounting either. I have tried to address these issues here without removing any functionality. I do, however, see a caveat with these changes, depending on whether users are using default values for fs_type and mountpath. There may very well be better ways to do this.

I do apologize if dragging two differing issues into one request is bad karma. My experience in these matters are rather limited for now.

### DIFF
--- a/manifests/logical_volume.pp
+++ b/manifests/logical_volume.pp
@@ -8,11 +8,11 @@ define lvm::logical_volume (
   $options           = 'defaults',
   $pass              = '2',
   $dump              = '1',
-  $fs_type           = undef,
+  $fs_type           = 'ext4',
   $mkfs_options      = undef,
-  $mountpath         = undef,
+  $mountpath         = "/${name}",
   $mountpath_require = false,
-  $mountlv           = false,
+  $createfs          = true,
   $extents           = undef,
   $stripes           = undef,
   $stripesize        = undef,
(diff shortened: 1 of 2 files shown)


tphoney:
Thanks for effort you have put in, and the explanation of the work you have carried out. Could you please look into the TravisCI failure.

TorLdre:
tphoney: Thank you for your feedback. I will redesign my proposal (I found the error).

TorLdre:
tphoney: I have done some more work. I have restored the default values (so the caveat described above is no longer a concern). In sum, my proposed changes will provide:

1) Ordering physical volume -> volume group
2) $size is no longer mandatory - $extents can be used instead
3) A new parameter $createfs is provided (defaults to true). If true, the module will behave like before. If false, no creation of file system of mounting will be done.

I have tested creating, deleting, with or without createfs, extents and size. It seems to work pretty well.

PR #102, opened 2015-04-17
add a fact that list all pvs in a vg

duritong:
sometimes it might be interesting to know which pvs are within the
existing vgs


tphoney:
Thanks for the work, this needs to be rebased.

duritong:
Here you go and sorry for closing and re-opening again.

If you merge it *now*, it won't need another rebase... ;-)

PR #105, opened 2015-04-22
Handle undef value in lvm::logical_volume

riton:
When an `undef` value is passed as `lvm::logical_volume` name, `mkfs` is called with unwanted command line like

```
Error: Execution of 'mkfs.ext4 /dev/rootvg/' returned 1: mke2fs 1.42.9 (28-Dec-2013)
/dev/rootvg/ is not a block special device.
Proceed anyway? (y,n)
Error: /Stage[main]/Main/Lvm::Logical_volume[undef]/Filesystem[/dev/rootvg/]/ensure: change from absent to present failed: Execution of 'mkfs.ext4 /dev/rootvg/' returned 1: mke2fs 1.42.9 (28-Dec-2013)
/dev/rootvg/ is not a block special device.
Proceed anyway? (y,n)
```

This does not break anything, but we're here protected by `mkfs` that asks for confirmation (this fails in batch mode).

This patch only introduces a stupid test to check that `$name` isn't  `undef`.

My software versions:
* `CentOS 7`
* `puppet v3.7.2`
* `ruby 2.0.0p353`

### DIFF
--- a/manifests/logical_volume.pp
+++ b/manifests/logical_volume.pp
@@ -22,6 +22,10 @@ define lvm::logical_volume (
 
   validate_bool($mountpath_require)
 
+  if ($name == undef) {
+    fail("lvm::logical_volume \$name can't be undefined")
+  }
+
   if $mountpath_require {
     Mount {
       require => File[$mountpath],
--- a/manifests/volume.pp
+++ b/manifests/volume.pp
@@ -61,6 +61,11 @@ define lvm::volume (
   $extents = undef,
   $initial_size = undef
 ) {
+
+  if ($name == undef) {
+    fail("lvm::volume \$name can't be undefined")
+  }
+
   case $ensure {
     #
     # Clean up the whole chain.
--- a/manifests/volume_group.pp
+++ b/manifests/volume_group.pp
@@ -8,6 +8,10 @@ define lvm::volume_group (
 
   validate_hash($logical_volumes)
 
+  if ($name == undef) {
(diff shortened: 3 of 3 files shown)


bmjen:
Thanks @riton! Is it possible to get a sample manifest that would produce this error?

mhaskel:
@riton I don't understand how this error could possibly happen, name is required :\

riton:
I know that `name`is required. I found this error with a manifest like:

```puppet
class myclass($lv_name)
{
    ::lvm::logical_volume { $lvname:
        ensure            => present,
        volume_group      => 'rootvg',
        size              => '1G',
        fs_type           => 'ext4',
        mountpath         => '/tmp',
        mountpath_require => false
    }
}
class { '::myclass': lv_name => 'foo' }
```
A simple typo in my variable name (`$lvname` instead of `$lv_name`) scares me when I saw that `mkfs` was called on `/dev/${vg_name}/`.

This could also happen if someone does not do proper _user input_ validation in its manifest like in:

```puppet
class myclass2($lv_name = undef) 
{
    ::lvm::logical_volume { $lv_name:
      [...]
     }
}
class { '::myclass2': }
```

I know that this is not supposed to happen, but considering that the final process is `mkfs` and that it _can_ possibly break things on the system, I think it's worth a simple test for undefined value.

bmjen:
@riton You are correct, a situation like the one you listed could definitely pass through an undef $name. I will merge.

mhaskel:
o.O

Good catch, @riton !

PR #108, opened 2015-05-16
Fix bad regexp

rabbitt:
regular expression was incorrectly using character classes and matching strings like: 100( or 100v etc.

### DIFF
--- a/lib/puppet/type/logical_volume.rb
+++ b/lib/puppet/type/logical_volume.rb
@@ -38,9 +38,9 @@ Puppet::Type.newtype(:logical_volume) do
   end
 
   newparam(:extents) do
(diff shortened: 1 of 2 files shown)


tphoney:
Thanks for spotting this, could you add a test to reflect the problem seen. This will prevent the issue from resurfacing.

rabbitt:
test added.

tphoney:
Sorry, i think the unit tests are failing.

rabbitt:
Sorry - I didn't realize you were supporting ruby 1.8.7. I'll switch to using hash rockets in the patch.

rabbitt:
What's the good word on this? :-)

tphoney:
Sorry about the delay, thanks for the work.

PR #89, opened 2014-12-04
Fixing Travis issues with PR#57 (Add swap support)

dgoldsmith:

PR #109, opened 2015-05-19
Add swap support.

ckoenig:
This pull request is based on https://github.com/puppetlabs/puppetlabs-lvm/pull/57 / https://github.com/puppetlabs/puppetlabs-lvm/pull/89, which wasn't merged.

It would be great, if you could have another look at this pull request.

### DIFF
--- a/manifests/logical_volume.pp
+++ b/manifests/logical_volume.pp
@@ -26,24 +26,41 @@ define lvm::logical_volume (
     fail("lvm::logical_volume \$name can't be undefined")
   }
 
-  if $mountpath_require {
+  $lvm_device_path = "/dev/${volume_group}/${name}"
+
+  if $mountpath_require and $fs_type != 'swap' {
     Mount {
       require => File[$mountpath],
     }
   }
 
-  $mount_ensure = $ensure ? {
-    'absent' => absent,
-    default  => mounted,
+  if $fs_type == 'swap' {
+    $mount_title     = $lvm_device_path
+    $fixed_mountpath = "swap_${lvm_device_path}"
+    $fixed_pass      = 0
+    $fixed_dump      = 0
+    $mount_ensure    = $ensure ? {
+      'absent' => absent,
+      default  => present,
+    }
+  } else {
+    $mount_title     = $mountpath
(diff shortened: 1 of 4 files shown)


mhaskel:
@ckoenig thanks for the contribution! Could you please squash this down to a single commit?

ckoenig:
No problem, now as a single commit.

manifests/logical_volume.pp
```diff
@@ -68,18 +85,38 @@
...
-      command => "mkdir -p ${mountpath}",
-      unless  => "test -d ${mountpath}",
-    } ->
-    mount { $mountpath:
+    if $fs_type == 'swap' {
+      if $ensure == 'present' {
+        exec { "swapon for '${mount_title}'":
+          path      => [ '/bin', '/usr/bin', '/sbin' ],
+          command   => "swapon ${lvm_device_path}",
+          unless    => "grep `readlink -f ${lvm_device_path}` /proc/swaps",
```
petems: Could use `swapon -s | grep ${mount_title}` for this?

ckoenig: Hm, sorry, but that doesn't really work in my case. It looks like this:
* ${lvm_device_path} = ${mount_title} = /dev/vg0/swap, which is a link to /dev/dm-2
* The output of `swapon -s` lists /dev/mapper/vg0-swap, which is also a link to /dev/dm-2
* The output of `cat /proc/swaps` lists /dev/dm-2

So in order to use the output of `swapon -s` we either have to extract the device and follow its redirect with `readlink -f` and do the same for ${lvm_device_path} to make the comparison work, or we adjust the way the device path is built. At the moment ${lvm_device_path} reflects the logic, which was used up until now and sets the path to /dev/${volume_group}/${name}. Instead of that we could probably use /dev/mapper/${volume_group}-${name}. That would make the comparison of the output of `swapon -s` easier.

Is there any specific problem with my approach, which doesn't occur with your proposed alternative?

petems: Makes sense :+1:

mhaskel:
:+1: thanks!

PR #110, opened 2015-05-20
added force parameter to physical_volume

crayfishx:
This PR adds a force parameter to the physical_volume resource type resulting in pvcreate being called with the --force flag.

This is particularly useful when the build process leaves disks with digital signatures which causes pvcreate to ask whether or not to wipe them, defaulting to no and failing the resource.  This was my use case and I'm using this patch in production as a workaround

### DIFF
--- a/lib/puppet/provider/physical_volume/lvm.rb
+++ b/lib/puppet/provider/physical_volume/lvm.rb
@@ -3,8 +3,12 @@ Puppet::Type.type(:physical_volume).provide(:lvm) do
 
     commands :pvcreate  => 'pvcreate', :pvremove => 'pvremove', :pvs => 'pvs', :vgs => 'vgs'
 
+    def force
+      @resource[:force] == :true ? '--force' : nil
+    end
+
     def create
-        pvcreate(@resource[:name])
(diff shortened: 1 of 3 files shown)


hunner:
Looks good, but needs a rebase.

tphoney:
@crayfishx  This needs rebased and the unit tests are failing for puppet 2.7

iankronquist:
@hunner Should I adopt this PR, or do you think @crayfishx is still interested in fixing it?

iankronquist:
It's been over two weeks so I'm going to hijack this patch. I move we close this PR and I'll open a new one up to date with the repo containing this patch.

iankronquist:
It turns out that this feature was actually reimplemented some time in late July.
This commit seems to do the same thing:
https://github.com/puppetlabs/puppetlabs-lvm/commit/c3e6b463033c46d7dfbb3b2f2329edbf6688ee1c

This should definitely be closed as fixed.

HelenCampbell:
Thank you @crayfishx for your work. Closing due to fix already being implemented as mentioned by @iankronquist (Thanks for you time spent looking into this!)

PR #112, opened 2015-06-10
(maint) Fixes specs for Puppet ~> 2.7.0

petems:
@DavidS - "puppet 2.7 calls all facts twice. recommended approach is to add '.at_least(1)' to expectations."


petems:
:+1:

PR #116, opened 2015-07-07
Support older versions of facter

spielkind:
Old facter versions do not support 'Facter::Util::Resolution.which' but they
are still used in LTS distributions like Ubuntu 12.04.

### DIFF
--- a/lib/facter/lvm_support.rb
+++ b/lib/facter/lvm_support.rb
@@ -4,8 +4,8 @@ Facter.add('lvm_support') do
   confine :kernel => :linux
 
   setcode do
-    vgdisplay = Facter::Util::Resolution.which('vgs')
-    vgdisplay.nil? ? nil : true
+    `which vgs`
+    $?.success? ? true : nil
   end
 end
 
--- a/spec/unit/facter/lvm_support_spec.rb
+++ b/spec/unit/facter/lvm_support_spec.rb
@@ -21,7 +21,6 @@ describe 'lvm_support fact' do
     context 'when vgs is absent' do
       it 'should be set to no' do
         Facter::Util::Resolution.stubs('exec') # All other calls
-        Facter::Util::Resolution.expects('which').with('vgs').at_least(1).returns(nil)
         Facter.value(:lvm_support).should be_nil
       end
     end


tphoney:
@spielkind thanks for the work, but the travis CI is failing. The PR cannot be merged until they pass.

spielkind:
@tphoney I've seen this, unfortunately I'm not that familiar with rspec, I'll look into that when I have some time, any help is appreciated.

iankronquist:
@spielkind are you interested in fixing this? I have experience with rspec and can help you if you want.

spielkind:
@iankronquist sure any help would be great.

lib/facter/lvm_support.rb
```diff
@@ -4,8 +4,8 @@
   confine :kernel => :linux
 
   setcode do
-    vgdisplay = Facter::Util::Resolution.which('vgs')
-    vgdisplay.nil? ? nil : true
+    `which vgs`
+    $?.success? ? true : nil
```
iankronquist: I noticed that the return value of the ternary operator are reverse of the original. Is this on purpose? [Ternary operators](http://alvinalexander.com/blog/post/ruby/examples-ruby-ternary-operator-true-false-syntax) are confusing sometimes.
This is what is causing one of the tests to fail.

lib/facter/lvm_support.rb
```diff
@@ -4,8 +4,8 @@
   confine :kernel => :linux
 
   setcode do
-    vgdisplay = Facter::Util::Resolution.which('vgs')
```
iankronquist: The other test is failing because this function is no longer being called. I think you need to remove line 24 of `spec/unit/facter/lvm_support_spec.rb`.

iankronquist:
@spielkind are you still interested in fixing this?

lib/facter/lvm_support.rb
spielkind: Hm, this should matter? In the original it's checked on nil and returns 'true' when not nil. The new one checks on success and returns 'true' when success ist true. Both cases return true under the same conditions?

tphoney:
I am closing this, as it is very stale. Please feel free to re-submit this with the necessary changes.

PR #117, opened 2015-07-13
Support mkfs_cmd - fixes MODULES-2215

robinbowes:
Fix for https://tickets.puppetlabs.com/browse/MODULES-2215


tphoney:
Thanks for the work, could you please squash your commits, some notes in the README would be great as well.

robinbowes:
Commits squashed.

I did consider udpating the README but it is so hopelessly out of date and not in any kids of standard format that I didn't bother.

PR #118, opened 2015-07-22
add minor device no and persist options

robinbowes:
I needed to create LVs with persistent block device IDs (for use a ext4 journal devices)

This change surfaces the --persistence and --minor options to the lvcreate command.


robinbowes:
Aw, crap. Let me clean that up.

PR #119, opened 2015-07-22
add minor device no and persist options

robinbowes:
I needed to create LVs with persistent block device IDs (for use a ext4 journal devices)

This change surfaces the --persistence and --minor options to the lvcreate command.


robinbowes:
Ticketed in https://tickets.puppetlabs.com/browse/MODULES-2261

lib/puppet/type/logical_volume.rb
```diff
@@ -46,6 +46,24 @@
...
+      unless value =~ /^[yn]$/i
+        raise ArgumentError , "#{value} is not a valid value for persistence"
+      end
+    end
+  end
+
+  newparam(:minor) do
+    desc "Set the minor number"
+    validate do |value|
+      unless value =~ /^(?:[0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])$/
```
hunner: How about `if value.to_i > 255 or value.to_i < 0` instead of a regex

bmjen:
A couple notes:

1. The new parameters need to be documented in the README and the examples 
2. The :persistent parameter definition should follow common idioms with a boolean value ( `true` or `false`), then the `y` or `n` values applied in the provider.

lib/puppet/type/logical_volume.rb
robinbowes: It's a fair cop, guv.

robinbowes:
Updated to address @hunner's comment, and @bmjen's 2nd point.

Regarding the README, I'd be happy to add the new params and some examples to the README if it were in some sort of sane format that made that relatively easy. However, it seems to me that the README needs a complete overhaul to bring it into line with the recommended standard layout. I think that is out of the scope of this PR.

abednarik:
Thi issue should be closed rigth https://tickets.puppetlabs.com/browse/MODULES-2261?

PR #120, opened 2015-07-31
add force option for physical_volume

timogoebel:
This PR adds the ability to pass the "force" option to pvcreate.


timogoebel:
Any news on this? What do you think?

timogoebel:
@tphoney : As you seem to have merged some things here and there are a couple of open prs... What do you think?

walkamongus:
I just stumbled across this conversation. I ran into this issue as well and submitted PR #135 as one way to fix.

PR #122, opened 2015-08-18
Adding option package management and mountpoint management

DevOpsFu:
Added some support for specifying a package name and version to be ensured against. The default is `lvm2` and `installed` if `manage_package` is set to true.

Added support for specifying mount points as a hash of `file` objects. The user only needs to specify a path to the intended mountpoint directory. Default owner/group and permissions can be overridden if required. 

The params defaults are set such that these changes should be non-breaking to existing users.

### DIFF
--- a/manifests/config_mountpoints.pp
+++ b/manifests/config_mountpoints.pp
@@ -1,4 +1,4 @@
-class lvm::config (
+class lvm::config_mountpoints (
   $mountpoints        = $::lvm::mountpoints,
 ) {
 
--- a/manifests/config_mountpoints.pp
+++ b/manifests/config_mountpoints.pp
@@ -1,0 +1,17 @@
+class lvm::config (
+  $mountpoints        = $::lvm::mountpoints,
+) {
+
+  if ($mountpoints) {
+
+    $mountpoint_defaults = {
+      ensure => directory,
+      mode   => '0755',
+      owner  => 'root',
+      group  => 'root',
+    }
+
+    create_resources(file, $mountpoints, $mountpoint_defaults)
+  }
+
+}
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -14,5 +14,5 @@ class lvm (
   class { '::lvm::install':             } ->
   class { '::lvm::config_mountpoints':  } ->
   class { '::lvm::config_volumes':      } ->
-  Class['Lvm']
+  Class['lvm']
 }
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -11,8 +11,11 @@ class lvm (
   validate_bool($manage_package)
   validate_string($package_name)
 
-  class { '::lvm::install':             } ->
-  class { '::lvm::config_mountpoints':  } ->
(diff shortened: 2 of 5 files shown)


manifests/install.pp
```diff
@@ -0,0 +1,15 @@
+class lvm::install (
+  $manage_package  = $::lvm::manage_package,
+  $package_name    = $::lvm::package_name,
+  $package_ensure  = $::lvm::package_ensure,
```
hunner: As a private class, this should not accept any parameters.

manifests/params.pp
```diff
@@ -0,0 +1,16 @@
...
+  case $::kernel {
+    'Linux': {
+      $manage_package = false
+      $package_name   = 'lvm2'
+      $package_ensure = 'installed'
+      $mountpoints    = {}
+      $volume_groups  = {}
+    }
+
+  default: {
```
hunner: Indentation

manifests/init.pp
```diff
@@ -1,10 +1,21 @@
...
 
   validate_hash($volume_groups)
+  validate_hash($mountpoints)
+  validate_bool($manage_package)
+  validate_string($package_name)
 
-  create_resources('lvm::volume_group', $volume_groups)
+  contain '::lvm::install'
+  contain '::lvm::config_mountpoints'
+  contain '::lvm::config_volumes'
```
hunner: Because this module must still support < 3.5, we can't use contain. Could you change this to the anchor pattern?

manifests/config_mountpoints.pp
```diff
@@ -0,0 +1,17 @@
...
+) {
+
+  if ($mountpoints) {
+
+    $mountpoint_defaults = {
+      ensure => directory,
+      mode   => '0755',
+      owner  => 'root',
+      group  => 'root',
+    }
```
mhaskel: I think these defaults are too permissive.

mhaskel:
Hi @DevOpsFu ,

Thanks for the work! In addition to the comments above, this will need to be squashed (should be a single commit) and you'll need to update the tests (https://github.com/puppetlabs/puppetlabs-lvm/blob/master/spec/unit/classes/lvm_spec.rb) to contain the facts you're depending on in params.pp https://github.com/puppetlabs/puppetlabs-apt/blob/master/spec/classes/apt_spec.rb#L3

DevOpsFu:
Hey @mhaskel and @hunner ,

Thanks for the feedback and comments, I'll try to get onto all of those items when time allows! :+1: 

Chris.

manifests/config_mountpoints.pp
DevOpsFu: @mhaskel what do you recommend?

underscorgan: For default values probably something more like 750 would be better. You can always add a parameter if those need to be more configurable.

tphoney:
Ping @DevOpsFu any luck on getting to this ?

bmjen:
@DevOpsFu Going to close this for now. Please reopen when you've had time to address @underscorgan and @hunner comments.

PR #124, opened 2015-10-12
Add parameter lvm_device_path

standagh:
While using multipath driver it is convenient to use different
device name - /dev/mapper/... instead /dev/\<volume_group\>/...
So variable lvm_device_path is moved to parameters with expected
value set as a default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,14 @@ Logical volume size can be extended, but not reduced -- this is for
 safety, as manual intervention is probably required for data
 migration, etc.
 
+### Size of the Logical Volumes
+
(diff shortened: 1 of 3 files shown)


iankronquist:
@standagh Thanks for your help! It looks like this will need a rebase, and that CI tests are failing. Specifically, you'll want to change line 199 in `./spec/unit/puppet/provider/logical_volume/lvm_spec.rb` to expect the new arguments you add in your PR.

standagh:
Pleas ignore this PR. There should be only single commit in PR.

PR #125, opened 2015-10-13
If size_is_minsize is true, consider the size in sync if already larg…

dfairhurst:
…er than desired size

Proposed fix for:
https://tickets.puppetlabs.com/browse/MODULES-2662

### DIFF
--- a/lib/puppet/provider/logical_volume/lvm.rb
+++ b/lib/puppet/provider/logical_volume/lvm.rb
@@ -134,20 +134,19 @@ Puppet::Type.type(:logical_volume).provide :lvm do
     end
 
     def size=(new_size)
-        lvm_size_units = { "K" => 1, "M" => 1024, "G" => 1048576, "T" => 1073741824, "P" => 1099511627776, "E" => 1125899906842624 }
-        lvm_size_units_match = lvm_size_units.keys().join('|')
+        lvm_size_units = { "K" => 1, "M" => 1024, "G" => 1024**2, "T" => 1024**3, "P" => 1024**4, "E" => 1024**5 }
 
         resizeable = false
         current_size = size()
 
-        if current_size =~ /(\d+\.{0,1}\d{0,2})(#{lvm_size_units_match})/i
+        if current_size =~ /^([0-9]+(\.[0-9]+)?)([KMGTPE])/i
             current_size_bytes = $1.to_f
-            current_size_unit  = $2.upcase
+            current_size_unit  = $3.upcase
         end
 
-        if new_size =~ /(\d+\.{0,1}\d{0,2})(#{lvm_size_units_match})/i
+        if new_size =~ /^([0-9]+(\.[0-9]+)?)([KMGTPE])/i
             new_size_bytes = $1.to_f
-            new_size_unit  = $2.upcase
+            new_size_unit  = $3.upcase
         end
 
         ## Get the extend size
--- a/lib/puppet/type/logical_volume.rb
+++ b/lib/puppet/type/logical_volume.rb
@@ -22,7 +22,7 @@ Puppet::Type.newtype(:logical_volume) do
   newparam(:initial_size) do
     desc "The initial size of the logical volume. This will only apply to newly-created volumes"
     validate do |value|
-      unless value =~ /^[0-9]+(\.[0-9]+)?[KMGTPEL]/i
+      unless value =~ /^[0-9]+(\.[0-9]+)?[KMGTPE]/i
         raise ArgumentError , "#{value} is not a valid logical volume size"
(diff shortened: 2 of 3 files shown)


lib/puppet/type/logical_volume.rb
```diff
@@ -35,6 +35,27 @@
         raise ArgumentError , "#{value} is not a valid logical volume size"
       end
     end
+    def insync?(is)
+        lvm_size_units = { "K" => 1, "M" => 1024, "G" => 1048576, "T" => 1073741824, "P" => 1099511627776, "E" => 1125899906842624 }
```
DavidS: By convention all code is two spaces indents.

For clarity, please use `1024*1024`, etc, instead of 1048576.

lib/puppet/type/logical_volume.rb
```diff
@@ -35,6 +35,27 @@
...
+        lvm_size_units_match = lvm_size_units.keys().join('|')
+
+        if is =~ /(\d+\.{0,1}\d{0,2})(#{lvm_size_units_match})/i
```
DavidS: I'd match a float with `\d+\.?\d*`, to avoid the `{}` ranges. It would still allow `1.g`, but `to_f` can parse that.

DavidS: Reading more code, this should use the same regex style as the validate function.

DavidS:
Congrats on getting all the tests green on your first try! The implementation looks fine in principle, I just nitpicked a few style issues. Please take care to use `git commit --amend` and `git push --force` when uploading your changes to this branch, to keep everything in one commit.

lib/puppet/type/logical_volume.rb
dfairhurst: Here I used pretty much the same code as is in provider/logical_volume/lvm.rb.
I'll see if both can be improved.  Not sure how to solve the duplication...

DavidS: I was looking at this: https://github.com/puppetlabs/puppetlabs-lvm/pull/125/files#diff-92d09665e3a40866889b747edead8e3fL34

I'm less concerned about duplication on this scale, as long as the regexes are consistent.

lib/puppet/type/logical_volume.rb
dfairhurst: I fixed the spacing, but I opted for 1024**2 etc. instead.

dfairhurst:
Updated as requested.  One thing that I could possibly change is to use \d instead of what was existing [0-9].
Also, I removed accepting the "L" as a size unit in the validate, since it does not seem to be supported in the provider and I couldn't find what this L could mean.  Let me know what you think about that.

lib/puppet/provider/logical_volume/lvm.rb
```diff
@@ -134,20 +134,19 @@ def size
...
+        lvm_size_units = { "K" => 1, "M" => 1024, "G" => 1024**2, "T" => 1024**3, "P" => 1024**4, "E" => 1024**5 }
 
         resizeable = false
         current_size = size()
 
-        if current_size =~ /(\d+\.{0,1}\d{0,2})(#{lvm_size_units_match})/i
+        if current_size =~ /^([0-9]+(\.[0-9]+)?)([KMGTPE])/i
             current_size_bytes = $1.to_f
-            current_size_unit  = $2.upcase
+            current_size_unit  = $3.upcase
```
DavidS: When tracking the unit handling code, I found another doozey in the overall code structure. Looking at https://github.com/puppetlabs/puppetlabs-lvm/pull/125/files#diff-d8b38c124d9da799f07a02cb607ba2deR125 , it uses lvs to convert the units, as lvm uses lowercase to denote multiples of 1024 and uppercase to denote multiples of 1000. o.O

See also http://linux.die.net/man/8/lvs --units.

dfairhurst: Is it really an issue if all the lvm commands called are passed with a lowercase unit, or just as bytes?

I don't think the module should have to support uppercase meaning multiples of 1000 instead of 1024.  It should always assume 1024 as long as it calls the correct lvm commands (i.e. don't use uppercase units when calling lvm)

DavidS: This code - and that is not your fault! - makes my brain hurt. Indeed in L122, incoming units are always downcased.

dfairhurst: :) anything else remaining or OK for merging?

DavidS:
Thanks for your contribution!

dfairhurst:
Thanks for the review + merge

PR #128, opened 2015-10-27
force ext4 filesystems so we can use whole disks

mrgum:
Hello

I use the filesystem resource to manage whole disks rather than partitions and mkfs.ext4 gives a y/n question unless the force (-F) flag is given.  this patch does that.

Thanks.

Neil

### DIFF
--- a/lib/puppet/provider/filesystem/lvm.rb
+++ b/lib/puppet/provider/filesystem/lvm.rb
@@ -22,7 +22,7 @@ Puppet::Type.type(:filesystem).provide :lvm do
     end
 
     def mkfs(fs_type)
-        mkfs_params = { "reiserfs" => "-q" }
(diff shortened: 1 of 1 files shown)


bmjen:
Thanks @mrgum, I'm not sure if this change is something we can safely introduce for all users. 

As stated in the mkfs.ext4 man page:
```
-F
Force mke2fs to create a filesystem, even if the specified device is not a partition on a block special device, or if other parameters do not make sense. In order to force mke2fs to create a filesystem even if the filesystem appears to be in use or is mounted (a truly dangerous thing to do), this option must be specified twice.
```

I think it would be preferable to have the force option be added as a parameter so that when a user wants to force the mkfs, that they explicitly pass that in vs making everyone that creates a ext4 filesystem have to use force.

mrgum:
life goes on, 8 years later lets move on

PR #131, opened 2015-11-04
(MODULES-3789) Lint the LVM Module's Puppet Code

iankronquist:
A simple Puppet lint pass. If you would like me to do rubocop that will be far more intensive. There are over 300 rubocop violations in the tests alone, and most of them should be fixed by hand.


HelenCampbell:
Thanks!

PR #136, opened 2015-12-04
Added support to the resize of a logical volume with swap

ricciocri:
Hello,

I noticed that the provider did not extended automatically the logical volume for a swap partition, so I've done this small change that follow the best practive for doing it.
swapoff (turn off swap)
mkswap (format the logical volume again, so it uses the whole size)
swapon (put swap back online).

Best Regards


walkamongus:
just ran into this recently as well -- would love this!

PR #138, opened 2015-12-16
sanitize vg names within fact names

duritong:
It is possible to name a VG something like vg-1, we should
downcase the - if we're going to use it as a name for facter.


jordanbreen28:
Closing as PR is outdated.

PR #139, opened 2015-12-17
Fix manage_pkg scope

mcanevet:

PR #140, opened 2015-12-23
$manage_pkg is a local variable

raphink:
and it's not a fact.


walkamongus:
#139 already has this.

raphink:
Ah good. I'll close and hope #139 will be merged soon.

PR #143, opened 2016-01-21
Set default mount option for dump to 0

chrw:
I would set the default for the dump mount option to 0. I'm not sure about the default with Redhat-based distributions but with Debian/Ubuntu the standard value seems to be 0. Is there a reason for the default value to be 1?


eputnam:
@chrw it doesn't look like there was a reason and Puppet's default is also 0. Great catch!

PR #147, opened 2016-02-04
Added the createonly option for the volumegroup and the possibility t…

ricciocri:
…o give more information for the PV.

In our infrastructure we have some highlevel configuration in hiera (our global.yaml) that is read via a hiera_hash (with deep merge) by all nodes, this is useful to have all the disk with the same setup.

But we have some exception and so to manage "easily" them we added the option createonly to the define  lvm::volume_group, while working on this we found useful to give more information about the physical volumes as well, so we added a new define and the possibility to use an hash to pass more information to the define lvm::volume_grop. 

Compatibility with the old configuration is mantained, so if physical_volume it's an array nothing change from the past behaviour.

Changed the README to show these changes as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,8 +104,33 @@ lvm::volume_groups:
         mountpath: /var/backups
         mountpath_require: true
 ```
-
-
+or to just build the VG if it does not exists
+```yaml
+---
+lvm::volume_groups:
+  myvg:
+    createonly: true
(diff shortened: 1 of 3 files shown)


gmarkey:
ETA on merge?

bmjen:
Thanks @ricciocri !

PR #149, opened 2016-03-15
Add mounted parameter to logical volume type

matthewrsj:
Mounted parameter added to logical_volume type. Ticket 3108.

Users should now be able to declare a `mounted` parameter as described in the README, like so:

```
logical_volume {'lv1':
  ensure => present,
  volume_group => 'myvg',
  mounted => true,
  size => '1G',
}
```

### DIFF
--- a/lib/puppet/type/logical_volume.rb
+++ b/lib/puppet/type/logical_volume.rb
@@ -163,6 +163,9 @@ Puppet::Type.newtype(:logical_volume) do
           end
       end
   end
+  newparam(:mounted) do
+      desc "Specifies if puppet should mount the volume. This only affects what puppet will do, and not what will be mounted at boot-time"
+  end
 
 
   autorequire(:volume_group) do
--- a/spec/unit/type/logical_volume_spec.rb
+++ b/spec/unit/type/logical_volume_spec.rb
@@ -430,4 +430,19 @@ describe Puppet::Type.type(:logical_volume) do
       )
     }.to_not raise_error
   end
+  
+  it 'mounted passed as a parameter does not throw an error' do
+    expect {
+      resource = Puppet::Type.type(:logical_volume).new(
+				{
+        :name             => 'fred',
+        :ensure           => :present,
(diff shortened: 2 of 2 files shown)


lib/puppet/type/logical_volume.rb
```diff
@@ -163,6 +163,9 @@ def insync?(is)
           end
       end
   end
+  newparam(:mounted) do
+      desc "Specifies if puppet should mount the volume. This only affects what puppet will do, and not what will be mounted at boot-time"
```
hunner: Could you make this 2 space indentation?

spec/unit/type/logical_volume_spec.rb
```diff
@@ -430,4 +430,19 @@
       )
     }.to_not raise_error
   end
+  
+  it 'mounted passed as a parameter does not throw an error' do
+    expect {
+      resource = Puppet::Type.type(:logical_volume).new(
+				{
```
hunner: And no hard tabs. Lots of times this is on the previous line, like `.new({`

hunner:
I see the `mounted` parameter on https://github.com/puppetlabs/puppetlabs-lvm/blob/master/manifests/logical_volume.pp#L15 and this ends up just getting passed to `ensure` of the `mount` resource and is not used by the declared `logical_volume` lvm type. I think there is confusion about what should be happening...

The readme documents the type `logical_volume` and claims that it has a parameter of `mounted`, but actually it's a defined type of `lvm::logical_volume` that actually has the parameter... I see `lvm::logical_volume`, `lvm::volume`, and `lvm::volume_group` defined types that have no documentation. It appears that the types `logical_volume`, `physical_volume`, `volume_group`, and `filesystem` types have documentation, but it may be conflated with the actual defined types and their parameters.

hunner:
So it's the readme that should actually be updated in this case, not the type.

hunner:
Thanks for your contribution; closing due to inactivity. Feel free to reopen and update the branch with the readme updates if you have time!

PR #153, opened 2016-05-20
add a device parameter, since we might want to use devicemapper

rockpapergoat:
we have a need to use device mapper multipathing on our storage, so it would be nice to move this variable to a param to allow overriding this behavior.

### DIFF
--- a/manifests/logical_volume.pp
+++ b/manifests/logical_volume.pp
@@ -11,6 +11,7 @@ define lvm::logical_volume (
   $fs_type           = 'ext4',
   $mkfs_options      = undef,
   $mountpath         = "/${name}",
+  $lvm_device_path   = "/dev/${volume_group}/${name}",
   $mountpath_require = false,
   $mounted           = true,
   $createfs          = true,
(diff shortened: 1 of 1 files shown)


manifests/logical_volume.pp
```diff
@@ -11,6 +11,7 @@
   $fs_type           = 'ext4',
   $mkfs_options      = undef,
   $mountpath         = "/${name}",
+  $lvm_device_path   = "/dev/${volume_group}/${name}",
```
hunner: I believe `$volume_group` will be unset at the time of parameter defaults as only `$name` is available in defaults. Could you migrate this to the $_volume_group pattern to make it work? Also, as no tests failed this is apparently not tested so a test would be great :).

The readme should mention the new parameter too. Thanks!

eputnam:
@rockpapergoat, thanks for opening this issue. It hasn't seen activity in a while, so I'm going to close it. This doesn't mean that we don't think this issue is valid, we're just trying to assess what issues are still relevant. If you still need this(even if you're not willing to work on it yourself) please feel free to reopen it. Again thanks for opening the issue, users who report bugs and ask for features are some of the most important players in the open source community. Feel free to hop on Slack (#forge-modules) or IRC (#puppet-dev, #voxpupuli) to chat further.

PR #154, opened 2016-06-17
Add support for thin provisioning and setting poolmetadatasize

afalko:


lib/puppet/provider/logical_volume/lvm.rb
```diff
@@ -110,7 +115,14 @@ def create
             args.push('--type', @resource[:type])
         end
 
-        args << @resource[:volume_group]
+	if @resource[:thin]
```
hunner: if `@resource[:thin]` is `:false` or `'false'` then this will be truthy. The type should either munge or only accept literal `true` and `false` values.

crayfishx:
@afalko : I've addressed the point made by @hunner , made some other tweaks and added tests and documentation.   I've raised this as a PR against your branch.... https://github.com/afalko/puppetlabs-lvm/pull/1 

... yep, a PR to a PR, github is sweet.

@afalko Can you review and merge the above PR and then it should make it through to this one - we are keen to see this PR merged.

Thanks!

crayfishx:
Before merging this, @afalko can you take a look at https://github.com/afalko/puppetlabs-lvm/pull/2 - I think we need this otherwise there is no way of configuring thinpools from hiera...etc.

lib/puppet/type/logical_volume.rb
```diff
@@ -74,6 +76,20 @@ def insync?(is)
     end
   end
 
+  newparam(:thin, :boolean => true, :parent => Puppet::Parameter::Boolean) do
+    desc "Set to true to create a thin provisioned logical volume"
+    defaultto false 
+  end
+
+  newparam(:poolmetadatasize) do
+    desc "Change the size of logical volume pool metadata"
```
vinzent: it is only used in the creat method. so it is not possible to change it.

lib/puppet/provider/logical_volume/lvm.rb
```diff
@@ -110,7 +118,12 @@ def create
             args.push('--type', @resource[:type])
         end
 
-        args << @resource[:volume_group]
+        if @resource[:thin]
+            args.push('--thin')
+            args << @resource[:volume_group] + "/" + @resource[:name]
```
vinzent: according `lvcreate --help` the syntax is `lvceate ... --thin VolumeGroupName/PoolLogicalVolume` the `:name` here is IMO the name of the lv to be created not the pool lv.

the `-n (--name)` is IMO still required and should be set to `:name`

lib/puppet/type/logical_volume.rb
```diff
@@ -74,6 +76,20 @@ def insync?(is)
...
+    validate do |value|
+      unless value =~ /^[0-9]+(\.[0-9]+)?[KMGTPE]/i
+        raise ArgumentError , "#{value} is not a valid logical volume size"
```
vinzent: wrong error text

crayfishx:
@vinzent your first point about `--extents`, you can use it, in fact that's how I tested it.... from the lvcreate manpage....

> Specifying the optional argument --size  or  --extents will  cause the creation of the thin pool logical volume.  Specifying the optional argument --virtualsize will cause the creation of the thin logical volume from given thin pool volume.   Specifying  both  arguments will cause the creation of both thin pool and thin volume using this pool.

lib/puppet/type/logical_volume.rb
crayfishx: Good point - `lvresize` does support this so it can be added easily enough

lib/puppet/type/logical_volume.rb
crayfishx: I'll change that, well spotted.

lib/puppet/provider/logical_volume/lvm.rb
crayfishx: I'm not sure about this..... I'll test it

vinzent: as you quoted from the man page. `--thin` seems to have multiple outcomes. but there is `--thinpool` too. for me it seems like `--thinpool` was introduced later to be specific what happens.

lib/puppet/type/logical_volume.rb
crayfishx: Summary from IRC discussion:  This is better documented and changed to a property in the future.

lib/puppet/type/logical_volume.rb
crayfishx: Changed in https://github.com/afalko/puppetlabs-lvm/pull/2

vinzent:
related: https://tickets.puppetlabs.com/browse/MODULES-2888

crayfishx:
@afalko @vinzent FYI

@vinzent and I had a long discussion on this over IRC - I'll summarize here...

The terminology and documentation around thin pools and thin pool volumes (which are different things but both are logical volumes) is confusing.   This PR enables the creation of _thin pools_.  What it doesn't do is allow you to create _thin pool volumes_ inside this thin pool.  For that reason, and after a discussion about some future work that will be raised in separate tickets, we've changed the attribute `thin` to `thinpool` to imply the creation of a thin pool.

So currently, to create a _thin pool_ you declare

```puppet
# Create a thin pool
logical_volume { 'blah':
  volume_group => 'foo',
  thinpool => true,
}
```

The logic behind this is that we would (in a future PR) introduce the `thin` parameter to  add functionality to create thin volumes within this pool, something like;

```puppet
# Create a thin pool *volume*
logical_volume { 'mythinvol':
  thin => true,
  pool => 'blah',
}
```

I think I've summarized that properly!

The changes reflecting this are in a PR to this PR (https://github.com/afalko/puppetlabs-lvm/pull/2) - if you are happy with the above @afalko could you please merge that into this PR?

We'll create issues separate to this PR capturing:

* Creation of thin pool _volumes_
* Changing `poolmetadatasize` into a configurable property rather than a parameter

afalko:
@vinzent @crayfishx : I've resolved the merge conflicts.

crayfishx:
Oops - minor mistake caught by travis - @afalko if you can update this PR with https://github.com/afalko/puppetlabs-lvm/pull/3 we should be good!

lib/puppet/type/logical_volume.rb
```diff
@@ -1,3 +1,5 @@
+require 'puppet/parameter/boolean'
```
hunner: This was added in puppet 3.3.0 so the version in metadata should be bumped.

README.md
```diff
@@ -167,12 +167,14 @@ resources out yourself.
...
 * persistent (Parameter) - Set to true to make the block device persistent
+* poolmetadatasize (Parameter) - Set the initial size of the logical volume pool metadata on creation
 * readahead (Parameter) - The readahead count to use for the new logical volume.
 * region_size (Parameter) - A mirror is divided into regions of this size (in MB), the mirror log uses this granularity to track which regions are in sync. CAN NOT BE CHANGED on already mirrored volume. Take your mirror size in terabytes and round up that number to the next power of 2, using that number as the -R argument.
 * size (Property) - The size of the logical volume. Set to undef to use all available space
 * size_is_minsize (Parameter) Default value: `:false` - Set to true if the ‘size’ parameter specified, is just the minimum size you need (if the LV found is larger then the size requests this is just logged not causing a FAIL)
 * stripes (Parameter) - The number of stripes to allocate for the new logical volume.
 * stripesize (Parameter) - The stripesize to use for the new logical volume.
+* thinpool (Parameter) - Default value: `:false` - Set to true to create a thin pool
```
hunner: s/:false/false/

hunner:
@crayfishx will fix in follow-up PR

unixsanet:
Just updated my environment to use LVM-0.8.0 and it appears to be incompatible with Satellite 6.1 and using Smart Class Parameter.  This code worked with 0.7.0:

```
oraclevg:
  physical_volumes:
  - /dev/sdb
  logical_volumes:
    lvd01:
      size: 24.99G
      mountpath: /d01
```

but fails with 0.8.0 with the following error:

`Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Invalid parameter thinpool on Logical_volume[lvd01] at /etc/puppet/environments/production/modules/lvm/manifests/logical_volume.pp:99`

vinzent:
@unixsanet you're probably facing the longstanding environment isolation problem with ruby types. See https://docs.puppet.com/puppet/latest/environment_isolation.html . If you have multiple environments with different versions of the same ruby types I'll can get into troubles like this.

walkamongus:
@unixsanet I ran into the exact same issue and a restart of puppetserver on the affected master resolved it.

tenajsystems:
I was following this PR and was not sure if the PR for creating thin pool volumes(thin-provisioned lv) like the example @crayfishx gave like below or #159. I do know that creating a thin pool was implemented in the PR but want to know if thin pool volumes was also implemented. Could someone please confirm this for me? 

    # Create a thin pool
    logical_volume { 'blah':
    volume_group => 'foo',
    thinpool => true,
    }

    # Create a thin pool *volume*
    logical_volume { 'mythinvol':
    thin => true,
    pool => 'blah',
    }

Thanks.

tenajsystems:
It appears the thinpool => true, doesn't work. I have setup the below and manually trying to create a thin LV using the command ```lvcreate -V 50GB -T vg-gluster/data -n Test``` and I get this message: ```Logical volume vg-gluster/data is not a thin pool```

    # Create a thin pool
    logical_volume { 'blah':
    volume_group => 'foo',
    thinpool => true,
    }

What I'm I doing wrong?

PR #156, opened 2016-07-29
Add timeout of 30 to execution statements of LVM facts

esalberg:
This is a followup to here:
https://github.com/puppetlabs/puppetlabs-lvm/pull/98

I figured it was worth putting in a new request versus reopening the old one.  I'd still like to get a timeout on a couple of the facts so that facter will return / doesn't hang on nodes with an underlying disk issue.

( [...] ,timeout: 30)

I set the timeout to 30 as a generic value, but I'm happy to adjust.  I still wasn't sure how to pass a value from the Puppet code into the custom fact, but if I can get a pointer on that, I'd be happy to make that change as well.


esalberg:
Closing due to testing - opening a new PR

PR #157, opened 2016-08-11
Update fact to add 30 second timeout; refactor due to Windows error

esalberg:
The confines started failing for the Windows test node after switching to Facter::Core::Execution.execute to add the timeout.  Rewrote the fact to fix the issues, including adding new array facts for vgs and pvs.

### DIFF
--- a/lib/facter/lvm_support.rb
+++ b/lib/facter/lvm_support.rb
@@ -5,7 +5,7 @@ Facter.add('lvm_support') do
 
   setcode do
     vgdisplay = Facter::Util::Resolution.which('vgs')
-    vgdisplay.nil? ? nil : true
+    vgdisplay.nil? ? nil : nil
   end
 end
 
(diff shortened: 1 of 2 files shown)


esalberg:
This is a followup to here:
#98

I figured it was worth putting in a new request versus reopening the old one. I'd still like to get a timeout on a couple of the facts so that facter will return / doesn't hang on nodes with an underlying disk issue.

( [...] ,timeout: 30)

I set the timeout to 30 as a generic value, but I'm happy to adjust. I still wasn't sure how to pass a value from the Puppet code into the custom fact, but if I can get a pointer on that, I'd be happy to make that change as well.

I also made some logic changes due to my test Windows node not liking the confine on the vgs command section (it tried to run that command).  Let me know if the Ruby can be improved, as that's not my forte.

esalberg:
Bear with me - I'm trying to fix the spec errors and then will squash all the commits.  Any suggestions would be appreciated.

esalberg:
Closing in favor of PR 158, which is a simpler change.

PR #158, opened 2016-08-22
Update fact to add 30 second timeout; add ifs to restrict lvm commands

esalberg:
This is a followup to here:
#98

I figured it was worth putting in a new request versus reopening the old one. I'd still like to get a timeout on a couple of the facts so that facter will return / doesn't hang on nodes with an underlying disk issue.

( [...] ,timeout: 30)

I set the timeout to 30 as a generic value, but I'm happy to adjust. I still wasn't sure how to pass a value from the Puppet code into the custom fact, but if I can get a pointer on that, I'd be happy to make that change as well.

I also added if statements around the lvm commands - because they are not in setcode blocks, the confines listed don't actually apply to them, and with Facter::Core::Execution being used to set the timeout, the error about the missing command now shows up (versus the different way it was handled with Facter::Util::Resolution).

Truthfully, the fact might benefit from being rewritten to be structured (at least add a Facter array for the vg_list and pv_list arrays that are used in the Ruby code) - I started down that path, but my Ruby / rspec isn't the strongest and I figured I'd do a quick patch for now and let someone else tackle a full refactor.

### DIFF
--- a/lib/facter/lvm_support.rb
+++ b/lib/facter/lvm_support.rb
@@ -14,7 +14,11 @@ end
 vg_list = []
 Facter.add('lvm_vgs') do
   confine :lvm_support => true
-  vgs = Facter::Util::Resolution.exec('vgs -o name --noheadings 2>/dev/null')
+
+  lvm = Facter.value(:lvm_support)
+  if lvm 
+    vgs = Facter::Core::Execution.execute('vgs -o name --noheadings 2>/dev/null', timeout: 30)
+  end
   if vgs.nil?
     setcode { 0 }
   else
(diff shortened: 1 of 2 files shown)


esalberg:
The rspec tests are still failing despite my attempts.  Help please?

tphoney:
Can you please rebase, this is why travisCI maybe failing for this PR. I tested this with #168  
apologies for the inconvenience.

esalberg:
Completely missed the update.  Rebased.

esalberg:
I am still having  rspec issues (same as above) - if someone could take a look and point me in the right direction, that would be great.

PR #163, opened 2016-11-07
[MODULES-3991] Issue with using /dev/disk/by-path/pci-XXXX:XX:XX.X-scsi-X:X:X:X 

timhughes:
Attempt to fix https://tickets.puppetlabs.com/browse/MODULES-3991

with the following code in create_vg.pp it physically works but I still get some output that is incorrect.

```
class { 'lvm':
  volume_groups    => {
    'myvg' => {
      physical_volumes => [ '/dev/disk/by-path/pci-0000:00:04.0-virtio-pci-virtio1', ],
      logical_volumes  => {
        'opt' => {
          'size'              => '200M',
          'mountpath'         => '/opt',
        },
      },
    },
  },
}
```

```
[root@localhost ~]# puppet apply create_vg.pp 
Warning: Config file /etc/puppet/hiera.yaml not found, using Hiera defaults
Notice: Compiled catalog for localhost.islab2.lmax in environment production in 0.20 seconds
Notice: /Stage[main]/Lvm/Lvm::Volume_group[myvg]/Volume_group[myvg]/physical_volumes: physical_volumes changed ['/dev/vdb'] to '/dev/disk/by-path/pci-0000:00:04.0-virtio-pci-virtio1'
Notice: Finished catalog run in 0.90 seconds
```

### DIFF
--- a/lib/puppet/provider/logical_volume/lvm.rb
+++ b/lib/puppet/provider/logical_volume/lvm.rb
@@ -77,7 +77,7 @@ Puppet::Type.type(:logical_volume).provide :lvm do
             args.push('--stripesize', @resource[:stripesize])
         end
 
-	
+
 
         if @resource[:poolmetadatasize]
             args.push('--poolmetadatasize', @resource[:poolmetadatasize])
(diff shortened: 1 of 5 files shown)


timhughes:
@lukebigum I think we need to do a bit of cleanup before we can get this upstream. I am going to close this request and move  the work into the company org

PR #173, opened 2016-12-09
Cannot add a new volume if the volume group has the same name. Fix: Better regexp for parsing the output of lvs

ruriky:
There is a bug which prevents user from adding a new logical volume if the volume name is the same as the volume group's name.

The regexp which parses the output of lvs command searches only for the name of the new logical volume. If the output contains previously added logical volumes, the regexp will match with the volume group's name of those volumes. Because the logical volume does not actually exist yet, the proceeding function which tries to extract the size of the new volume will fail.

This fix extends the regexp to match not only with the logical volume name put with the combination of logical volume name and the volume group name.

Example output of lvs command:
```
  LV   VG   Attr       LSize Pool Origin Data%  Meta%  Move Log Cpy%Sync Convert
  comp data -wi-ao---- 5.00g
  root data -wi-ao---- 5.00g
  swap data -wi-ao---- 4.00g
```

Example of failing conf:
```
---
lvm::volume_groups:
  data:
    physical_volumes:
      - /dev/sda2
      - /dev/sda3
    logical_volumes:
      comp:
        size: 5G   
      data:
        size: 5G
      root:
        size: 5G

```

### DIFF
--- a/lib/puppet/provider/logical_volume/lvm.rb
+++ b/lib/puppet/provider/logical_volume/lvm.rb
@@ -293,7 +293,9 @@ Puppet::Type.type(:logical_volume).provide :lvm do
     private
 
     def lvs_pattern
-        /\s+#{Regexp.quote @resource[:name]}\s+/
+        # lvs output format:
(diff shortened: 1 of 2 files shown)


bmjen:
Thanks @ruriky !

PR #174, opened 2016-12-13
[#puppethack] fix syntax error in volume.pp

swart000:
This commit fixes issue MODULES-4172.
Syntax error (using "+" with string) in fail message of volume.pp.
+ replaced with sprintf for concatanation error string.

### DIFF
--- a/manifests/volume.pp
+++ b/manifests/volume.pp
@@ -137,7 +137,7 @@ define lvm::volume (
 
     }
     default: {
-      fail ( sprintf("%s%s", 'puppet-lvm::volume: ensure parameter can only ',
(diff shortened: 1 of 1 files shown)


manifests/volume.pp
```diff
@@ -137,8 +137,8 @@
 
     }
     default: {
-      fail ( 'puppet-lvm::volume: ensure parameter can only be set to ' +
-        'cleaned, absent or present' )
+      fail ( sprintf("%s%s", 'puppet-lvm::volume: ensure parameter can only ',
```
eputnam: Rake is throwing a fit but locally the only thing lint complains about is the double quotes here. If you could just change those to single I think we'll be good to go.

PR #179, opened 2017-01-17
Fix Error when blkid returns no TYPE field

ashkop:
If device has no file system yet, `blkid` return no TYPE information in its output on Ubuntu 16.04. This results is the following error output in Puppet:
```
Debug: Executing: '/sbin/blkid /dev/sdc1'
Error: /Stage[main]/Filesystem[/dev/sdc1]: Could not evaluate: undefined method `[]' for nil:NilClass
```

### DIFF
--- a/lib/puppet/provider/filesystem/lvm.rb
+++ b/lib/puppet/provider/filesystem/lvm.rb
@@ -16,7 +16,8 @@ Puppet::Type.type(:filesystem).provide :lvm do
     end
 
     def fstype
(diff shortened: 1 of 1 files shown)


mciurcio:
There is typo: fstype_match VS fs_type_match

csm10495:
Can something like this be merged? I just hit this issue when attempting to mount from a nonstandard location (like /dev/disk/<stuff>)

cedws:
Bumping this, it's still an issue and am having to use an Exec resource as a workaround.

PR #181, opened 2017-01-26
added (missing) followsymlinks parameter

a-dekker:
Like addressed in https://tickets.puppetlabs.com/browse/MODULES-4284, I am missing this parameter when using the new "by-path" functionality.


eputnam:
@a-dekker , thanks for the PR!

PR #182, opened 2017-03-08
Setting thinpool default to false

missingcharacter:
According to docs https://github.com/puppetlabs/puppetlabs-lvm/blob/master/README.md, thinpool is default to false, but as we see in https://github.com/puppetlabs/puppetlabs-lvm/blob/master/manifests/logical_volume.pp#L24 it is default to undef and then it is always called in https://github.com/puppetlabs/puppetlabs-lvm/blob/master/manifests/logical_volume.pp#L92


eputnam:
@missingcharacter thanks for the PR! Great catch.

PR #184, opened 2017-03-28
replace validate_* with datatypes

bastelfreak:


bastelfreak:
The goal of this is to replace all the legacy validate_* functions, not to introduce perfect datatypes for all parameters.

PR #185, opened 2017-04-10
Removing the 'extent size' check in logical_volume provider

lukebigum:
There are several situations where Puppet will not increase the size of an LV but if you run the equivalent LVM command it will actually work. See [MODULES-4180](https://tickets.puppetlabs.com/browse/MODULES-4180).

Also changes the test to not expect such a specific Puppet error. Finally some minor whitespace and lint fixes.


eputnam:
@lukebigum thanks for the PR!

PR #187, opened 2017-05-23
puppethack - MODULES-4753 and MODULES-4964

dhollinger:
The exec resources running swapon and swapoff in the lvm::logical_volume
define type were running whenever a swap logical volume was created
or updated.

The logical_volume provider was also running these same commands on a swap
resize. This caused a conflict where the puppet run would error out
when the logical_volume provider attempted to run swapoff on a swap
lvm that had already been unloaded from swap and the lvm would never
be reloaded into swap.

When the exec resources were removed from the lvm::logical_volume
defined type, it created a gap wherein a NEW swap lvm would not be
loaded into swap after being created.

To solve this issue, I added the swapon command and logic that will
run that command against the filesystem resource if the fs_type parameter
is set to 'swap'

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 .*.swp
 Gemfile.lock
 .bundle/
+.fixtures/modules
+.fixtures/manifests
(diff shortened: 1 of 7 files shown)


dhollinger:
Needs updated tests. Still working on that.

dhollinger:
Updated with the capability to remove Swap LVMs

PR #189, opened 2017-07-10
Added Hash as acceptable type for physical_volumes

stivesso:
The line 11 checks if is_hash and that condition can't apply if the only acceptable type are Array and String
The error I got and that is fixed is:
parameter 'physical_volumes' expects a value of type Array or String, got Struct at ...


uvNikita:
I'm also getting the same error when trying to configure `physical_volumes` parameters using hiera syntax. Is there something that's blocking the merge of this fix?

PR #191, opened 2017-07-28
AIX updates for 7.1

spidersddd:
This update is to add some error checking, and get the AIX implementation of LVM more inline with Linux distros.  I have added some AIX params that are needed for the new tooling and AIX functionality. 

These changes should not change the functionality for Linux base systems.


jordanbreen28:
Closing as PR is outdated.
Please re-open if necessary.

PR #194, opened 2017-10-24
Add support for resizeing with extents

scornelissen85:
I have created an option to resize the logical volume by setting the extents. To use this option, the new parameter resize_extents must be set to true.

It cannot be merged with master currently because I have build this change upon version 0.9.0 as this is the latest version currently out. And we use still Puppet 3 (which isn't working anymore on the master). Is it possible to create a 0.9.0 or a Puppet-3 branch?


tenajsystems:
What's the status of this pull request? I am looking to extending my logical volume and this feature would be a great help.

PR #198, opened 2017-11-23
MODULES-6065 - Resolving facter error when lvm not installed

pmcmaw:
Currently the code is trying to execute before the 'lvm_support' fact has been resolved therefore code that should not be executed is. Adding setcode to around the block ensures that the whole block of code has been executed before being used. This has happens in 2 code blocks therefore adding the change to both.


pmcmaw:
Similar PR merged.

PR #204, opened 2018-02-01
Add possbility to overwrite $lvm_device_path

dn1s:
this is neccessary to bindmount.
Also updated if statement for mount to be
able to bindmount.


dn1s:
I do understand that the main purpose of this manifest is to create and manage logical_volumes and volume_groups. But since you do a mount if the logical_volume does not exist I thought that we could have the feature to use bind mount with this module.

PR #206, opened 2018-03-02
(MODULES-6449) remove misleading confinement

vchepkov:
confine statement applies only to 'setcode' block


vchepkov:
confine statements are misleading, but 'tests' expect them to be there, oh well.

PR #209, opened 2018-06-11
move code to setcode to apply confine

raemer:
Code requiring lvm commands was moved into the setcode block as the
confine statement is not applied to this code otherwise, causing the
facter run to fail on systems where these commands are not available
(i.e. OpenBSD).


jst-cyr:
@raemer : My apologies that you never got a response to your PR in the past. I'm reviewing our stale and open PRs and after 8 years of being open I'm recommending we cancel/close this PR and reopen in the future if the need arises for somebody again.

PR #211, opened 2018-07-04
Added mountpath file resourcce

Akhilez:
Error "Could not find dependency File[$mountpath]" is solved with this change.


vchepkov:
The idea was to define resource elsewhere and then set mountpath_require to true

Akhilez:
You're right. I was using it wrong. Thanks!

PR #213, opened 2018-08-17
Add tasks and `puppet resource` support for logical_volume

dylanratcliffe:
The PR adds the following functionality:

  - `puppet resource logical_volume` now works
  - Tests now pass
  - Adds tasks for common LVM operations

I wrote the tasks in ruby so as to take advantage of the types and providers and to avoid duplicating logic. New features and bugfixes in the types and providers should automatically apply to the tasks too. The disadvantage of this approach is that the tasks will not work unless run on a machine that has both Puppet installed and the `lvm` module on it (likely via pluginsync). This means they will not be useful for most bolt users, sorry bolt users.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ Gemfile.lock
 .bundle/
 .fixtures/modules
 .fixtures/manifests
+spec/fixtures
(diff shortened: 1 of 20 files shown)


dylanratcliffe:
Closing in favour of merging with #214

PR #215, opened 2018-08-30
Add facts, functions, tasks, plans

dylanratcliffe:
It's a big one.

## New features

  - Structured facts
  - Functions for converting to and from bytes
  - Tasks for common actions that use the types and providers
  - Expand plan for one-touch expanding
  - Modified provider to allow `puppet resource` to work for `logical_volumes`

## Fixes

  - Fixed issues that were causing compilation to fail in tests
  - Removed some redundant code in facts

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,6 +287,40 @@ parameters documented above also apply to AIX systems.
 * range (Parameter) - Sets the inter-physical volume allocation policy. AIX only
 * type (Parameter) - Configures the logical volume type. AIX only
 
+## Functions
+
+`lvm::bytes_to_size`: Converts a number of bytes to a size format that LVM can understand e.g. `lvm::bytes_to_size(214748364800)` would return `200g`
+
+`lvm::size_to_bytes`: Converts an LVM size to bytes, the opposite of `lvm::bytes_to_size`.
+
+## Tasks
+
+Documented in the ["Tasks" tab](https://forge.puppet.com/puppetlabs/lvm/tasks) on the forge
+
+## Plans
+
+### `lvm::expand`
+
+Executes common tasks for expanding mounts, this can include:
+
+* Creating PVs
+* Adding PVs to VG
+* Extending LV
+* Extending filesystem
+
+**Parameters:**
+
+`server`: *String* The target for the plan
+
+`volume_group`: *String* The volume group to which the logical volume belongs
+
+`logical_volume`: *String* The logical volume which is to be expanded
+
+`additional_size`: *String* How much size to add to the LV. This should be specified in LVM format i.e. "200m" or "2.5g"
+
+`disks`: *Array[String]* Any physical disks that should be added to the volume group as part of the expand process
+
+`resize_fs`: *Boolean* Whether or not to resize the filesystem
 
 ## Limitations
 
--- a/lib/puppet/type/logical_volume.rb
+++ b/lib/puppet/type/logical_volume.rb
@@ -13,9 +13,19 @@ Puppet::Type.newtype(:logical_volume) do
         raise ArgumentError, "Volume names must be entirely unqualified"
       end
     end
+
+    # This is a bit of a hack, drags the volume_group up from the provider if
+    # it was specified, allowing `puppet resource` and other things that rely
+    # on an indirector search to work
+    munge do |name|
+      if @resource.original_parameters[:provider].respond_to? :volume_group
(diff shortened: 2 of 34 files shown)


README.md
```diff
@@ -287,6 +287,40 @@ parameters documented above also apply to AIX systems.
 * range (Parameter) - Sets the inter-physical volume allocation policy. AIX only
 * type (Parameter) - Configures the logical volume type. AIX only
 
+## Functions
+
+`lvm::bytes_to_size`: Converts a number of bytes to a size format that LVM can understand e.g. `lvm::bytes_to_size(214748364800)` would return `200g`
```
tphoney: Generally we do not document functions in the readme, this can be removed

lib/puppet/type/logical_volume.rb
```diff
@@ -13,9 +13,19 @@
         raise ArgumentError, "Volume names must be entirely unqualified"
       end
     end
+
+    # This is a bit of a hack, drags the volume_group up from the provider if
+    # it was specified, allowing `puppet resource` and otrher things that rely
```
tphoney: typo

lib/puppet/type/logical_volume.rb
```diff
@@ -13,9 +13,19 @@
...
+    munge do |name|
+      if @resource.original_parameters[:provider].respond_to? :volume_group
+        @resource[:volume_group] = @resource.original_parameters[:provider].volume_group
+      end
+      name
+    end
   end
 
-  newparam(:volume_group) do
+  newproperty(:volume_group) do
```
tphoney: Changing this, will cause issues. It may be backwards incompatible

README.md
dylanratcliffe: Why wouldn't we document them? There aren't documented anywhere else are they?

lib/puppet/type/logical_volume.rb
dylanratcliffe: The problem with this being a param is that it's not possible to pull a param from `puppet resource` which will then cause all the tasks to fail. By definition this should have always been a property as it is able to be read back from the system and isn't just something that is provided on creation. Can we run the beaker suite or something to verify what issues it might create as this change is critical to this entire PR (except the facts)

dylanratcliffe: @tphoney Any feedback on this?

tphoney:
👍  from me. I am just going to get another review, due to its size

pmcmaw:
Hey @dylanratcliffe 

Something you may want to be aware of is that this module is currently only testing against puppet 4 in travis. I created a PR to enable puppet 5 and puppet6 however I am seeing quite a few failures. Bolt is only supported on ruby 2.3+ (2.3 is shipped with puppet4). Therefore if this PR make use of bolt in any way the configuration is not supported. 

This is the PR I have mentioned regarding updating the testing in Travis: #216

dylanratcliffe:
@pmcmaw This PR doesn't make use of any of Bolt's libraries at all so we shouldn't (and aren't) seeing any issues with Puppet 4. All changes to the T&Ps, functions and tests are Puppet 4 compatible as we've seen in the tests and should be fine with 5 and 6 also but we'll see once that PR is merged. This PR doesn't been to be blocked by #216 does it?

pmcmaw:
@dylanratcliffe no doesn't block just something I want to ensure you were aware of. 
Going to merge this, what a great PR! 👍

Tardja:
I didn't find how to post comment to code. Maybe I can't do it as PR already merged.
I noticed that changes in https://github.com/puppetlabs/puppetlabs-lvm/commit/aa5bb5cfc158ec473a2e67d9719632cc59ae2154 related to lib/facter/lvm_support.rb returned error that was fixed in PR 193:
`Error: Facter: error while resolving custom facts in /var/lib/puppet/lib/facter/lvm_support.rb: execution of command "vgs -o name --noheadings 2>/dev/null" failed: command not found.`

dylanratcliffe:
@Tardja I looked at that and thought the merge had worked correctly but maybe not, do you mean that the the error has now come back?

vchepkov:
@dylanratcliffe , yes, code now fails on systems without LVM binaries installed.
'confine' method only affects 'setcode' method, so any code outside of 'setcode' gets executed
I put comments in an existing ticket https://tickets.puppetlabs.com/browse/MODULES-6449

PR #216, opened 2018-10-22
(maint) - Adding puppet 5 and puppet 6 testing in travis

pmcmaw:
Currently travis is only running on 2 ruby version against puppet4. I have updated to run tests on the supported versions of puppet which are 4, 5 and 6 alongside the ruby version they are shipped with.


pmcmaw:
No longer required. Closing in favour of @ #223

PR #193, opened 2017-09-12
Don't execute the lvm commands when not supported.

smithj4:
When lvm_support was not present, facter would still attempt to execute the vgs and pvs commands.

PR #220, opened 2018-11-10
[WIP] Move command execution to within setcode blocks

dylanratcliffe:
This allows for confine statments to work correctly

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -1,5 +1,7 @@
 fixtures:
   repositories:
     "stdlib": "https://github.com/puppetlabs/puppetlabs-stdlib.git"
+  forge_modules:
+    mount_core: "puppetlabs/mount_core"
   symlinks:
     "lvm": "#{source_dir}"
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,13 +20,9 @@ matrix:
       env: PUPPET_GEM_VERSION="~> 4.0" STRICT_VARIABLES="yes"
     - rvm: 2.3.1
       env: PUPPET_GEM_VERSION="~> 4.0" STRICT_VARIABLES="yes"
-    - rvm: 2.1.6
+    - rvm: 2.4.4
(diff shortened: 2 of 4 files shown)


dylanratcliffe:
@vchepkov can you please review? RE: [MODULES-6449](https://tickets.puppetlabs.com/browse/MODULES-6449)

vchepkov:
Shall I wait till you finish the test suite? Also curious, what change now require dependency on mount_core ?

dylanratcliffe:
Sorry this last commit should have it fixed. Puppet 6 removes a lot of stuff from Puppet core and moves it out to modules like all the nagios types for example. Mount was also one of them so when upgrade to Puppet 6 you'll need to use the mount module instead of relying on the built in types

vchepkov:
First check, some facts disappeared, not sure if they were useful to begin with, but you asked to test

```
# diff -u before after|grep ^-
--- before	2018-11-10 12:37:55.796311076 +0000
-lvm_pv_0 => /dev/sda2
-lvm_vg_0 => vg0
-lvm_vg_vg0_pvs => /dev/sda2
```

vchepkov:
confinement works fine though. I tested on Solaris, on Linux without lvm

vchepkov:
It appears you call Facter.add from within of Facter.add. Is it allowed?

dylanratcliffe:
Hmm, tests are passing but that doesn't necessarily mean it's working. Possibly Facter.add from within a fact fails silently. Stand by while I refactor more

dylanratcliffe:
@vchepkov Yes all the stuff that was ripped out has been included in the agent packages for now so that everything doesn't totally fall over. I think this might be something to do with cfacter working differently than ruby facter. Trying to work out how to get a proper test suite set up so that I can verify

dylanratcliffe:
Closing in favour of the simpler #193 and the eventual removal of these facts

PR #222, opened 2018-11-20
Don't execute the lvm commands when not supported. #193

jonny0303:
Unfortunately you've deleted the changes, merged in https://github.com/puppetlabs/puppetlabs-lvm/pull/193 with the new release 1.1.0.

Error: Facter: error while resolving custom facts in /opt/puppetlabs/puppet/cache/lib/facter/lvm_support.rb: execution of command "vgs -o name --noheadings 2>/dev/null" failed: command not found.

### DIFF
--- a/lib/facter/lvm_support.rb
+++ b/lib/facter/lvm_support.rb
@@ -15,8 +15,10 @@ vg_list = []
 Facter.add('lvm_vgs') do
   confine :lvm_support => true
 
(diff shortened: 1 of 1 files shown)


vchepkov:
@dylanratcliffe is working on a fix in #220 

IMHO, these facts should be obsoleted, new structured facts serve the same purpose in more modern way

dylanratcliffe:
I vote we just merge this and do a separate PR for removing those facts with the next major release. Trying to make them work properly is barely worth it give that they are super old-school anyway

PR #224, opened 2018-12-06
(maint) Removed test code that missed review

dylanratcliffe:
Don't know how we missed this one, whoops.


tphoney:
Good catch @dylanratcliffe

PR #230, opened 2019-02-22
Fix false positive detection of prefixed lvs

rpluem:
If a logical volume with the name prefixlvname is created before the logical volume with name lvname then lvname is never created because exists? returns true for it. See also #231.

### DIFF
--- a/lib/puppet/provider/logical_volume/lvm.rb
+++ b/lib/puppet/provider/logical_volume/lvm.rb
@@ -157,7 +157,7 @@ Puppet::Type.type(:logical_volume).provide :lvm do
   end
 
   def exists?
-    lvs(@resource[:volume_group]) =~ %r{#{@resource[:name]}}
+    lvs(@resource[:volume_group]) =~ lvs_pattern
   rescue Puppet::ExecutionFailure
     # lvs fails if we give it an empty volume group name, as would
     # happen if we were running `puppet resource`. This should be
--- a/spec/unit/puppet/provider/logical_volume/lvm_spec.rb
+++ b/spec/unit/puppet/provider/logical_volume/lvm_spec.rb
@@ -34,6 +34,12 @@ describe provider_class do
       @provider.class.stubs(:lvs).with('data').returns(lvs_output)
       expect(@provider.exists?).to be > 10
     end
+    it "returns 'nil', lv 'swap' in vg 'VolGroup' exists" do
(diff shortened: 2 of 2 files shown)


rpluem:
A CLA for me (rpluem) should be on file with cla.puppet.com

rpluem:
Anything I can do to move this forward?

sheenaajay:
@rpluem Could you please add some tests to verify the change.

rpluem:
@sheenaajay Would you like to got with #230 or with #231 as solution for this issue? I have opened both PRs as possible solutions as I was not sure which one is seen as better. Once the solution is chosen I can add tests to the remaining one.

genebean:
@rpluem Can you provide a sample or two of things this PR or #231 would address?

rpluem:
Let's say you have a logical volume 'data1' in the volume group VG1 already created and you want to create a logical volume 'data' (notice 'data' is a substring of 'data1') in the volume group VG1 it wouldn't work because the current code of exists? would report the resource as already existing as it only checks if there are logical volumes that **contain** the string 'data' but not if there **is** a logical volume named 'data'. This is a regression and worked fine until dbad8025dcbaf6f69983e8d348efdbcdad807c61.

genebean:
I got an example via slack and am validating some ideas... hoping to post again shortly

genebean:
I adapted the code in this PR to have tests in https://github.com/puppetlabs/puppetlabs-lvm/pull/244

sheenaajay:
Thank you @genebean @rpluem .Will remove the following PR https://github.com/puppetlabs/puppetlabs-lvm/pull/244

rpluem:
I added a slightly different test compared to #244. It uses a logical volume name that is a substring of an existing logical volume name and not a substring of the volume group name which this patch fixes as well. So I think both tests should be in the final merge.

genebean:
The test I added in #244 was to directly address an error case a user in the community slack was facing. I used the scenario they were working around as validation so it might make sense to add both tests. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

rpluem:
And this is a good test since it checks for a scenario I did not have on my mind when creating the PR :-). It also shows that this pull request here is the solution as #231 would still fail your case. Hence I will close it.

rpluem:
Anything I can provide to move this forward?

rpluem:
This has been superseded by #244. Hence closing this one.

PR #231, opened 2019-02-22
 Fix false positive detection of prefixed lvs

rpluem:
If a logical volume with the name prefixlvname is created before the logical volume with name lvname then lvname is never created because exists? returns true for it. See also #230.


rpluem:
A CLA for me (rpluem) should be on file with cla.puppet.com.

rpluem:
Anything I can do to move this forward?

rpluem:
Closed in favour of #230 which provides a more complete solution that also works for the test case added in #244.

PR #243, opened 2020-01-16
correctly match lv name in lvs output

SimonPe:
The lv name is the first element in the table, so we should match only on that.
The inexact match resulted in `exists?` returning `true` when the lv name matches part of either the vg name or part of another lv name


SimonPe:
closing this, #230 should be the better way of fixing this (using `lvs_pattern`) I should have read trough open PR's before flying into the code.

PR #248, opened 2020-02-12
Post-1.3.0 mergeback

genebean:
Due to the mergeback not being done earlier master had to be merged into release to true things up w/o rebasing and, thus, changing the 1.3.0 tag.


genebean:
Self-merging as there are no actual changes

PR #251, opened 2020-03-23
Feature/add support for filesystem labels

unki:
_e.g._

```yaml
    logical_volumes:
      opt:
        size: 20G
        use_fs_label: true
```

will then lead to

* formating the filesystem by `mkfs.ext4 -L opt /dev/vg-xyz/opt`. the most `mkfs`'s implement the `-L` label-parameter. So it shouldn't be necessary to make it configurable.
* _/etc/fstab_ entry `LABEL=opt /opt ext4 defaults 0 0`

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,12 +142,12 @@ If you need a more complex configuration, you'll need to build the
 resources out yourself.
 
 ## Optional Values
-  The `unless_vg` (physical_volume) and `createonly` (volume_group) will check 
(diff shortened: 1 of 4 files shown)


unki:
failing tests are afaics not related to my changes.
possibly gets fixed by #260

jst-cyr:
@unki , my apologies that there has been no response previously on your PR. I'm doing a full review of all the open PRs across our modules, found this one, and am wondering if this is still a feature that you need in your setup?

PR #253, opened 2020-04-06
MODULES-10620:  Don't erase filesystem content and add force arg

ahmet2mir:
Hello,

When a device already contains a FS, the command mkfs use "force" and erase the content. It could be very dangerous causing data lost.

To avoid this, add a "force" arg to filesystem.
When set to default false

    a device without FS will be created
    a device with a FS will not be changed until user specify force, in this case umount and create the FS


jst-cyr:
My apologies, @ahmet2mir, that this PR has been sitting open for 6 years with no approvals. The CLA has also not been signed. At this point, I think it's best to close this one out and if the need arises again we can look to reopen it then.

PR #257, opened 2020-08-03
workaround for removing an empty (no fs) logical volume

cbudziszewski:
blkid returns error code 2 if a path does not exist, which causes puppet to bail and not remove the lv.

A similar issue was already fixed for resizing (https://github.com/puppetlabs/puppetlabs-lvm/pull/188/files).

It may be better to wrap the blkid call somehow, since it is used multiple times in the code and has more than one valid exit code.

### DIFF
--- a/lib/puppet/provider/logical_volume/lvm.rb
+++ b/lib/puppet/provider/logical_volume/lvm.rb
@@ -149,7 +149,7 @@ Puppet::Type.type(:logical_volume).provide :lvm do
 
   def destroy
     name_escaped = "#{@resource[:volume_group].gsub('-', '--')}-#{@resource[:name].gsub('-', '--')}"
-    if File.exist?(path) and blkid(path) =~ %r{\bTYPE=\"(swap)\"}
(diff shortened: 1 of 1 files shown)


lib/puppet/provider/logical_volume/lvm.rb
```diff
@@ -149,10 +149,10 @@ def create
 
   def destroy
     name_escaped = "#{@resource[:volume_group].gsub('-', '--')}-#{@resource[:name].gsub('-', '--')}"
-    if blkid(path) =~ %r{\bTYPE=\"(swap)\"}
+    if File.exist?(path) and blkid(path) =~ %r{\bTYPE=\"(swap)\"}
```
sanfrancrisko, line 152:
```suggestion
    if File.exist?(path) && blkid(path) =~ %r{\bTYPE=\"(swap)\"}
```

cbudziszewski:
I did some more testing. turns out, this fix does not actually fix the problem. here is a new pull request for a better fix: https://github.com/puppetlabs/puppetlabs-lvm/pull/258

PR #258, opened 2020-08-04
blkid returning 2 is not an error in this case

cbudziszewski:
blkid returns error code 2 if a path does not exist, which causes puppet to bail and not remove the lv.

A similar issue was already fixed for resizing (https://github.com/puppetlabs/puppetlabs-lvm/pull/188/files).

It may be better to wrap the blkid call somehow, since it is used multiple times in the code and has more than one valid exit code.


jst-cyr:
Hello @cbudziszewski , I'm so sorry that there were no responses to your PR. I'm reviewing our open PRs and found this one from over 6 years ago. Given the age of the PR, I'm going to cancel this for now but we can always reopen in the future if it comes up again as a need for someone.

PR #260, opened 2021-01-11
MODULES-10915: use facter 3 for spec test

b4ldr:
This likley needs fixing in pdk or .sync but wanted to at least test the
fix first


b4ldr:
`lvm_support` spec test work with facter 3.11 all the others work with facter 4.  i can't see the difference thats triggering this

PR #261, opened 2021-04-10
(MODULES-11011) extend vg facts

atsonkov:
Adds facts for the extends in the volume groups (size, count, free extends)
Also sets the units to megabytes (of 1024k)
Otherwise the --nosuffix option was not used, and we could get results like '<54,24g'

Ref: MODULES-11011

### DIFF
--- a/lib/facter/volume_groups.rb
+++ b/lib/facter/volume_groups.rb
@@ -20,9 +20,12 @@ Facter.add(:volume_groups) do
       'vg_allocation_policy',
       'vg_size',
       'vg_free',
+      'vg_extent_size',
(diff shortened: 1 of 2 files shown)


atsonkov:
There are 7 CI tests failing for puppet 6, but the same tests are alos failing on master (not affected vy my changes).

Opened separate issue for fixing master (MODULES-11016).

jst-cyr:
Good day @atsonkov, I'm sorry your PR hasn't gotten any activity. In reviewing old PRs, I found this one. No activity after 5 years leads me to want to close this for now. If a need comes up in the future, let's reopen and revisit.

PR #263, opened 2021-04-24
size_to_bytes failes when output of lvs contains <

Rthur:
Example of lvs output where previous regular expression fails:

$ sudo lvs
  LV   VG   Attr       LSize   Pool Origin Data%  Meta%  Move Log Cpy%Sync Convert
  root rhel -wi-ao---- <21.50g
  swap rhel -wi-ao----   2.50g


Rthur:
resubmitting with email matching CLA

PR #265, opened 2021-09-22
Relax stdlib maximum version

jovandeginste:


dhs-rec:
#269 has already been merged which is a duplicate of this one, isn't it?

jovandeginste:
It would seem so. One wonders why the other PR was merged easily.

PR #266, opened 2021-12-15
Allow stdlib v8

sazzle2611:


sazzle2611:
Sorry but I'm totally confused, I don't understand how the change in stdlib version is causing the Travis CI errors?

I want to start contributing to puppet modules, I have used puppet for years and we use GitLab CI pipelines but I'm not familiar with Ruby or Travis.

Any info greatly appreciated!!

Thanks,
Sarah

towo:
> Sorry but I'm totally confused, I don't understand how the change in stdlib version is causing the Travis CI errors?

If you look at the other PRs, it seems to be more of a problem with the tests than with change you made. It just appears that this module is a bit abandoned (probably also because Puppet Inc really wants their contributors to jump through hoop and open JIRA tickets for PRs)

bastelfreak:
thanks for the PR, I merged the same thing in https://github.com/puppetlabs/puppetlabs-lvm/pull/269

PR #268, opened 2022-01-19
Delete empty common.yaml

bastelfreak:
Empty yaml files will create warnings in the Puppetserver log.


bastelfreak:
I'm going forward to merge this. with the broken yaml file, puppetserver will log a warning in every module call which generates a ton of noise for users.

PR #269, opened 2022-02-09
Bump stdlib dependency to v9

towo:
stdlib v9 is easily supported as there are no breaking changes relating to anything but Puppet/OS compatiblity.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -10,7 +10,7 @@
   "dependencies": [
     {
       "name": "puppetlabs/stdlib",
-      "version_requirement": ">= 4.13.1 < 7.0.0"
+      "version_requirement": ">= 4.13.1 < 9.0.0"
     },
     {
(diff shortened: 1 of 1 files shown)


towo:
Tests are failing independently of this change.

bastelfreak:
thanks for the PR. Merging this since CI is broken with no relation to the code change.

dhs-rec:
Any chance this will be released sometime? It's the only module I use that still prevents a stdlib upgrade (without `--ignore-dependencies`).

OH, and BTW: Could issues please be enabled on this repo (see https://tickets.puppetlabs.com/browse/FORGE-643)?

bastelfreak:
sadly I cannot do releases/cannot reenable issues :( @binford2k is that something you can take a look? there wasn't a pdksync in a while so the whole CI setup is missing (and I don't know how it should be configured).

PR #273, opened 2022-08-24
updating to new CI workflows

binford2k:


.travis.yml
```diff
@@ -1,4 +1,5 @@
 ---
+os: linux
```
bastelfreak, line 2: @binford2k should we delete the file?

bastelfreak (approved):

.travis.yml
binford2k: probably so 😁

PR #275, opened 2022-08-25
puppet-lint and rubocop cleanup

bastelfreak:


bastelfreak:
I think this is ready for merge. afterwards we can take a look at the rubocop failures and the failing other jobs

Issue #278, opened 2022-10-25
Wiping of signatures


SHWETHABHAT1:
## Describe the Bug
Currently the lvm2 is updated, as a result, lvcreate command required manual confirmation of signature wiping. But we dont have support to provide --yes option, during logical volume creation, as we dont have handle to provide options for lvcreate.

## Expected Behavior
Either provide a options, for --yes, or make yes option as default for prompting.

## Steps to Reproduce
1. Create a logical volume for first time.
2. Remove/Delete the volume
3. Try to create logical volume again and mount it, will throw error.

## Environment
lvm2-2.03.14-3.el8.x86_64.rpm
in rocky linux 8.6

bschonec:
I'll second this one as I ran into this problem today on a system that had an XFS logical volume deleted (not managed by puppet) and then when puppet went to create a new filesystem, I got "WARNING: xfs signature detected on /dev/vgdb/mariadb".  I had to "dd" the first few Kbytes of the space on the logical volume to get past the problem.

SHWETHABHAT1:
Is this repo still maintained ? Anyone working on this ? I don't see any activity

PR #279, opened 2022-11-03
Add yes argument when creating/extending a LV

bmlit:
# issue
(related to #278)
This module is failing when trying to create a XFS LV on location where a XFS LV was present :
```
Execution of '/sbin/lvcreate -n foo --size 160G main_vg' returned 5: WARNING: xfs signature detected on /dev/main_vg/foo at offset 0. Wipe it? [y/n]: [n]
Aborted wiping of xfs.
1 existing signature left on the device.
Failed to wipe signatures on logical volume main_vg/foo.
Aborting. Failed to wipe start of new LV.
```
# solution
A fix is to add the `--yes` option which can be used for automatically answer the above interactive question.

# implementation choices
This argument apply on the `lvcreate`/`lvextend` command as they're taking *new blocks* for the the LV, and therefore may trigger the issue.

I set it by default to `true` for being more versatile and avoiding Puppet run failure.
The man page discourage it (`Use  with  extreme  caution. (For automatic no, see -qq.)`) but Puppet is here for automating thinks and the user should know the consequences.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -67,7 +67,7 @@ Metrics/ParameterLists:
 # Offense count: 4
 # Configuration parameters: AllowedMethods, AllowedPatterns.
 Metrics/PerceivedComplexity:
-  Max: 33
+  Max: 35
 
 # Offense count: 3
 Naming/AccessorMethodName:


manifests/logical_volume.pp
```diff
@@ -28,6 +28,7 @@
   $no_sync                           = undef,
   $region_size                       = undef,
   $alloc                             = undef,
+  Boolean $yes                       = true,
```
bastelfreak, line 31: I'm wondering if this would be a breaking change because it changes the default behaviour. maybe it makes sense to default to false?

bastelfreak: I think it makes sense to merge this with false as default value and then create a new PR that changes it to true. but we will keep that on hold for the next major release.

SHWETHABHAT1:
If we set it to false by default, then is there no option to wipe signatures, or for sending --yes to the prompt ?

jordanbreen28:
@bmlit thanks for this! 
This module has recently become officially supported by the CAT Team at Puppet, so we are looking to get this moving forward!
Are you able to resolve the conflicts?

jordanbreen28:
Hey @bmlit - did you get the chance to review the conflicts?

jordanbreen28:
Nice @bmlit thanks for that!
Yeah I suspect this is due to recent changes to access to repo level variables for forked PRs.
Let me see if I can dig into it abit more and get this moving.

tasks/ensure_lv.json
```diff
@@ -80,6 +80,10 @@
     "region_size": {
       "description": "A mirror is divided into regions of this size (in MB), the mirror log uses this granularity to track which regions are in sync. CAN NOT BE CHANGED on already mirrored volume. Take your mirror size in terabytes and round up that number to the next power of 2, using that number as the -R argument.",
       "type": "Optional[Integer]"
-    } 
+    },
+    "yes_flag": {
+      "description": "Do not prompt for confirmation interactively but always assume the answer yes.",
```
jordanbreen28, line 85: Think we will need to update these docs to reflect the current release, then when default is set to true in the next major release we can change this back.

manifests/logical_volume.pp
```diff
@@ -78,6 +78,7 @@
   Optional[Boolean] $no_sync                                                    = undef,
   Optional[Variant[String[1], Integer]] $region_size                            = undef,
   Optional[Enum['anywhere', 'contiguous', 'cling', 'inherit', 'normal']] $alloc = undef,
+  Optional[Boolean] $yes_flag                                                   = false,
```
jordanbreen28, line 81: `$yes_flag` should never be undef, so we can remove the `Optional` portion
```suggestion
  Boolean $yes_flag                                                   = false,
```

README.md
```diff
@@ -201,6 +201,7 @@ resources out yourself.
 * stripesize (Parameter) - The stripesize to use for the new logical volume.
 * thinpool (Parameter) - Default value: `false` - Set to true to create a thin pool or to pool name to create thin volume
 * volume_group (Parameter) - The volume group name associated with this logical volume. This will automatically set this volume group as a dependency, but it must be defined elsewhere using the volume_group resource type.
+* yes_flag (Parameter) - Default value: `false` - Do not prompt for confirmation interactively but always assume the answer yes.
```
jordanbreen28, line 204: Same here, needs updated to reflect the current code :)

manifests/logical_volume.pp
jordanbreen28, line 81:
```suggestion
  Boolean $yes_flag                                                   = false,
```
Should be OK. `$yes_flag` won't ever be undef. :)

README.md
jordanbreen28, line 204:
```suggestion
* yes_flag (Parameter) - Default value: `false` - Do not prompt for confirmation interactively but always assume the answer false.
```
Same here, needs updated to reflect the current code :)

tasks/ensure_lv.json
jordanbreen28, line 85:
```suggestion
      "description": "Do not prompt for confirmation interactively but always assume the answer false.",
```
Think we will need to update these docs to reflect the current release, then when default is set to true in the next major release we can change this back.

May be better to stick with the boolean terms (true/false) as well to avoid confusion. :)

jordanbreen28: Agreed! That’s much better

jordanbreen28:
@bmlit can you rebase off of main?

jordanbreen28:
@bmlit couple of puppet-lint errors, could you look to resolve?

jordanbreen28 (approved):
LGTM.
Great work on this one @bmlit!

tr0k:
@jordanbreen28 when do you expect the next release (tag) for this contribution to be reflected?

jordanbreen28:
@tr0k we are just waiting on a couple of decisions to be made on other open PRs and bug fixes to go in before we release.
A release will be cut within the next couple of weeks. :-)

tr0k:
Thanks for the quick reply! W8 impatiently for any updates on this. Cheers!

croddy:
Will this be merged?

PR #283, opened 2023-05-16
(CONT-974) - Add puppet module support

jordanbreen28:


.fixtures.yml
```diff
@@ -1,10 +1,12 @@
...
   repositories:
-    'facts': 'git://github.com/puppetlabs/puppetlabs-facts.git'
-    'provision': "git://github.com/puppetlabs/provision.git"
-    'puppet_agent': 'git://github.com/puppetlabs/puppetlabs-puppet_agent.git'
-    'stdlib': "https://github.com/puppetlabs/puppetlabs-stdlib.git"
+    facts: 'https://github.com/puppetlabs/puppetlabs-facts.git'
+    provision: "https://github.com/puppetlabs/provision.git"
+    puppet_agent: 
+      repo: 'https://github.com/puppetlabs/puppetlabs-puppet_agent.git'
+      ref: v4.13.0
```
bastelfreak, line 7: why pin this module to a specific version?

.fixtures.yml
```diff
@@ -1,10 +1,12 @@
...
+    stdlib: "https://github.com/puppetlabs/puppetlabs-stdlib.git"
   forge_modules:
-    'mount_core': "puppetlabs/mount_core"
+    mount_core: "puppetlabs/mount_core"
   symlinks:
```
bastelfreak, line 11: puppetlabs_spec_helper will create the symlink, we don't need to define it

.fixtures.yml
```diff
@@ -1,10 +1,12 @@
...
-    'stdlib': "https://github.com/puppetlabs/puppetlabs-stdlib.git"
+    facts: 'https://github.com/puppetlabs/puppetlabs-facts.git'
+    provision: "https://github.com/puppetlabs/provision.git"
+    puppet_agent: 
+      repo: 'https://github.com/puppetlabs/puppetlabs-puppet_agent.git'
+      ref: v4.13.0
+    stdlib: "https://github.com/puppetlabs/puppetlabs-stdlib.git"
   forge_modules:
-    'mount_core': "puppetlabs/mount_core"
+    mount_core: "puppetlabs/mount_core"
```
bastelfreak, line 10: I think it makes sense to switch to the git repo here as well

CHANGELOG.md
```diff
@@ -1,253 +1,274 @@
-# Change log
+<!-- markdownlint-disable MD024 -->
+# Changelog
 
-All notable changes to this project will be documented in this file. The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) and this project adheres to [Semantic Versioning](http://semver.org).
+All notable changes to this project will be documented in this file.
```
bastelfreak, line 4: was the changelog rewrite intentional?

data/common.yaml
```diff
@@ -0,0 +1 @@
+--- {}
```
bastelfreak, line 1: at vox pupuli we avoid empty hiera files, I'm not sure if CAT/Puppet has a best practice here.

lib/facter/lvm_support.rb
```diff
@@ -11,34 +13,31 @@
...
 #   Number of VGs
-vg_list = []
 Facter.add('lvm_vgs') do
   confine lvm_support: true
 
-  if Facter.value(:lvm_support)
-    vgs = Facter::Core::Execution.execute('vgs -o name --noheadings 2>/dev/null', timeout: 30)
-  end
+  setcode do
+    vgs = Facter::Core::Execution.execute('vgs -o name --noheadings 2>/dev/null', timeout: 30) if Facter.value(:lvm_support)
```
bastelfreak, line 20: I think the condition isn't required because the `lvm_vgs` fact is confined on `lvm_support`
```suggestion
    vgs = Facter::Core::Execution.execute('vgs -o name --noheadings 2>/dev/null', timeout: 30) if Facter.value(:lvm_support)
```

lib/puppet/provider/filesystem/aix.rb
```diff
@@ -161,7 +160,7 @@ def attribute_flag(pvalue)
       size: 'size',
       extended_attributes: 'ea',
       mount_options: 'options',
-      encrypted: 'efs',
+      encrypted: 'efs'
```
bastelfreak, line 163: I'm not a fan of this.  I think the vox pupuli rubocop config enforces a comma after the last element. If you now add a new element on the next line, you need to modify the previous line as well. That slightly increases the diff and can make rebases and reverts really tricky/ugly.

manifests/init.pp
```diff
@@ -1,14 +1,18 @@
 # == Class: lvm
```
bastelfreak, line 1: the old comment can also be deleted to please puppet-strings
```suggestion

```

manifests/logical_volume.pp
```diff
@@ -1,35 +1,85 @@
 # == Define: lvm::logical_volume
```
bastelfreak, line 1:
```suggestion

```

manifests/logical_volume.pp
```diff
@@ -1,35 +1,85 @@
...
-  $type                              = undef,
-  Variant[Boolean, String] $thinpool = false,
-  $poolmetadatasize                  = undef,
-  $mirror                            = undef,
-  $mirrorlog                         = undef,
-  $no_sync                           = undef,
-  $region_size                       = undef,
-  $alloc                             = undef,
+  String[1] $volume_group,
+  Optional[String[1]] $size                                                     = undef,
```
bastelfreak, line 56: I didn't dig through the whole code, but maybe it makes sense to allow Integer for size

manifests/physical_volume.pp
```diff
@@ -1,19 +1,22 @@
 # == Define: lvm::physical_volume
```
bastelfreak, line 1:
```suggestion
```

manifests/volume.pp
```diff
@@ -1,25 +1,38 @@
 # == Define: lvm::volume
```
bastelfreak, line 1:
```suggestion
```

.fixtures.yml
jordanbreen28: We opted to pin to v4.13.0 across our modules as it was the latest stable release that included puppet 8, and this is consistent and inline with the rest of our supported modules.
(and we preferred this approach rather than pulling from main)

manifests/volume.pp
```diff
@@ -1,25 +1,38 @@
...
 #
-# This defined type will create a <code>logical_volume</code> with the name of
-# the define and ensure a <code>physical_volume</code>,
-# <code>volume_group</code>, and <code>filesystem</code> resource have been
+# This defined type will create a logical_volume with the name of
+# the define and ensure a physical_volume,
+# volume_group, and filesystem resource have been
 # created on the block device supplied.
 #
 # === Parameters
```
bastelfreak, line 8:
```suggestion
```

.fixtures.yml
jordanbreen28: Updated. 👍

CHANGELOG.md
jordanbreen28: Yes this is auto-generated by our new changelog generator 💯

manifests/volume_group.pp
```diff
@@ -1,4 +1,19 @@
 # == Define: lvm::volume_group
+# @param physical_volumes The list of physical volumes to be included in the volume group; 
+# this will automatically set these as dependencies, but they must be defined elsewhere using the physical_volume resource type.
+
+# @param createonly Default value: false - If set to true the volume group will be created if it does not exist.
```
bastelfreak, line 5: puppet-strings will pick up the datatype and default value from the code, so you dont need to list it here.

metadata.json
```diff
@@ -19,37 +19,56 @@
   ],
   "operatingsystem_support": [
     {
-      "operatingsystem": "RedHat"
+      "operatingsystem": "RedHat",
+      "operatingsystemrelease": [
```
bastelfreak, line 23: the whole PR has a lot of changes. I suggest to pull some of them out into their own PRs. For example this metadata.json adjustment. It drops support for multiple OSses and also adds new ones if I read the diff correctly. I think that deserve its own PR.

metadata.json
```diff
@@ -19,37 +19,56 @@
...
+        "7.1",
+        "7.2"
+      ]
     }
   ],
   "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">= 4.6.1 < 7.0.0"
+      "version_requirement": ">= 7.0.0 < 9.0.0"
```
bastelfreak, line 68: same here, I think dropping multiple major versions should be a dedicated PR

jordanbreen28: Agreed, same as above.

.fixtures.yml
bastelfreak: Why do you prefer pinning for this module, but not for the rest? Usually we try to run unit tests against master/main and pinned versions from metadata.json for acceptance tests.

lib/facter/lvm_support.rb
jordanbreen28: Introduced as a bugfix in https://github.com/puppetlabs/puppetlabs-lvm/pull/222/files.
Interestingly this is confined `lvm_support: true`, so shouldn't execute the block in the first place.

.sync.yml
```diff
@@ -20,12 +20,11 @@ Rakefile:
...
+  unmanaged: true
+.github/workflows/nightly.yml:
+  unmanaged: true
 .github/workflows/release.yml:
   unmanaged: false
-.github/workflows/pr_test.yml:
-  unmanaged: false
-.github/workflows/spec.yml:
+.github/workflows/release_prep.yml:
   unmanaged: false
```
bastelfreak, line 30: I wanted to suggest to enable the puppet-lint plugins parameter_documentation / parameter_types but I'm currently not sure if/how pdk-templates supports that or if they are enabled by default.

manifests/logical_volume.pp
jordanbreen28: `size` is the size of the logical volume. It is passed in the format `20M`, `5G`, `15G`, (you can see some more details [here](https://forge.puppet.com/modules/puppetlabs/lvm/readme) in the readme)

.sync.yml
jordanbreen28: As mentioned these checks are enabled by default 💯

manifests/init.pp
jordanbreen28: Updated

manifests/physical_volume.pp
jordanbreen28: Updated

manifests/volume.pp
jordanbreen28: Updated

manifests/volume_group.pp
jordanbreen28: Updated

.fixtures.yml
jordanbreen28: So we've had instances in our not so distant past where breaking changes were introduced in `main` of puppet_agent, and that turned our ci into a sea of red.
This is one way we've tried to limit that from occurring. 💯

metadata.json
jordanbreen28: Created a PR for this https://github.com/puppetlabs/puppetlabs-lvm/pull/284

lib/facter/lvm_support.rb
```diff
@@ -48,20 +47,21 @@
...
 
-  if pvs.nil?
-    setcode { 0 }
-  else
-    pv_list = pvs.split
-    setcode { pv_list.length }
-  end
-end
+  setcode do
+    pvs = Facter::Core::Execution.execute('pvs -o name --noheadings 2>/dev/null', timeout: 30) if Facter.value(:lvm_support)
```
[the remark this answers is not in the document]
jordanbreen28: see above

lib/puppet/provider/filesystem/aix.rb
jordanbreen28: Thanks @bastelfreak and totally understood. The CAT Teams practice is to fix all non-breaking rubocop violations, and this is a couple we regularly enforce across our modules.

jordanbreen28:
removing backwards-incompatible label as the dropping of puppet 6 has now been included in a separate PR.

PR #284, opened 2023-05-30
(maint) - Add puppet 8/Drop puppet 6

jordanbreen28:


kenyon:
PR title should be updated to say that this added Puppet 7 support too, so that it appears in the changelog.

Issue #295, opened 2023-06-07
VG as absolute path


ic248:
Hi,

While submitting the other bug report I noticed that the data type for vg in lvm::volume is now Stdlib::Absolutepath. As far as I'm aware a volume group is just a string not a path? Below is an example of what we currently set for a test server and it will error out on vg.

`Evaluation Error: Error while evaluating a Resource Statement, Lvm::Volume[reports]: parameter 'vg' expects a Stdlib::Absolutepath = Variant[Stdlib::Windowspath = Pattern[/\A(([a-zA-Z]:[\\\/])|([\\\/][\\\/][^\\\/]+[\\\/][^\\\/]+)|([\\\/][\\\/]\?[\\\/][^\\\/]+)).*\z/], Stdlib::Unixpath = Pattern[/\A\/([^\n\/\0]+\/*)*\z/]] value, got String`

```
lvm::volume:
    reports:
        ensure: present
        vg: 'vg_reports'
        pv: '/dev/sdb'
        fstype: 'xfs'
        size: '200G'
```

The example in Puppet Forge (below) also shows it as a string.

`lvm::volume { 'lv_example0': vg => 'vg_example0', pv => '/dev/sdd1', fstype => 'ext4', size => '100GB', }`

Please say if you require any more info.

Thanks
Ian

jordanbreen28:
@ic248 thank you!!
Yes you're right, $vg should have a datatype of `String[1]`, surprised this wasn't picked up in testing..

PR #296, opened 2023-06-07
(GH-295) - fix vg datatype

jordanbreen28:
$vg in lvm::volume was incorrectly set to `stdlib::absolutepath`, when Volume group can take the format of String.

Fixes https://github.com/puppetlabs/puppetlabs-lvm/issues/295


manifests/volume.pp
```diff
@@ -64,7 +64,7 @@
 define lvm::volume (
   Enum['present', 'absent', 'cleaned'] $ensure,
   Stdlib::Absolutepath $pv,
-  Stdlib::Absolutepath $vg,
+  String[1] $vg,
```
bastelfreak, line 67: I suggest to add a unit test here that explicitly validates strings, otherwise people might change it back to `Stdlib::Absolutepath` during the next refactoring.

jordanbreen28: Agreed! Currently implementing, this should not have got past testing :)

bastelfreak (approved):

chelnak (approved):

PR #308, opened 2023-07-12
fix issues on nodes without lvm

### LINKED ISSUE #307: v2.0.3 is broken
## Describe the Bug
Updating to v2.0.3 causes failures:

```
2023-07-11 13:55:15 +0000 Facter (err): error while resolving custom facts in /opt/puppetlabs/puppet/cache/lib/facter/lvm_support.rb undefined method `message' for #<String:0x00000000047ec120>
```

Agent is RHEL8, lvm is not installed

rtib:
tested manually on nodes with and without lvm

Closes #307

### DIFF
--- a/README.md
+++ b/README.md
@@ -346,6 +346,13 @@ Logical volume size can be extended, but not reduced -- this is for
 safety, as manual intervention is probably required for data
 migration, etc.
 
+## Deprecation Notice
+
+Some facts reported by this module are being deprecated in favor of upcomming structured facts.  The following facts are being deprecated:
+
+* `lvm_vg_*`
+* `lvm_vg_*_pvs`
+* `lvm_pv_*`
 
 # Contributors
 
--- a/lib/facter/lvm_support.rb
+++ b/lib/facter/lvm_support.rb
@@ -25,8 +25,8 @@ Facter.add('lvm_vgs') do
   setcode { vg_list.length }
 end
 
-# # lvm_vg_[0-9]+
-# #   VG name by index
+# lvm_vg_[0-9]+
+#   VG name by index
 vg_list.each_with_index do |vg, i|
   Facter.add("lvm_vg_#{i}") do
     setcode { vg }
@@ -41,8 +41,8 @@ vg_list.each_with_index do |vg, i|
   end
 end
 
-# # lvm_pvs: [0-9]+
-# #   Number of PVs
+# lvm_pvs: [0-9]+
+#   Number of PVs
 pv_list = []
 Facter.add('lvm_pvs') do
   confine lvm_support: true
@@ -55,8 +55,8 @@ Facter.add('lvm_pvs') do
   setcode { pv_list.length }
 end
 
-# # lvm_pv_[0-9]+
-# #   PV name by index
+# lvm_pv_[0-9]+
+#   PV name by index
 pv_list.each_with_index do |pv, i|
   Facter.add("lvm_pv_#{i}") do
     setcode { pv }
--- a/lib/facter/lvm_support.rb
+++ b/lib/facter/lvm_support.rb
@@ -17,47 +17,48 @@ vg_list = []
 Facter.add('lvm_vgs') do
   confine lvm_support: true
 
-  vgs = Facter::Core::Execution.execute('vgs -o name --noheadings 2>/dev/null', timeout: 30)
-
-  if vgs.nil?
-    setcode { 0 }
-  else
-    vg_list = vgs.split
-    setcode { vg_list.length }
+  if Facter.value(:lvm_support)
+    vgs = Facter::Core::Execution.execute('vgs -o name --noheadings 2>/dev/null', timeout: 30)
+    vg_list = vgs.split unless vgs.nil?
   end
+
+  setcode { vg_list.length }
 end
 
-# lvm_vg_[0-9]+
-#   VG name by index
+# # lvm_vg_[0-9]+
+# #   VG name by index
 vg_list.each_with_index do |vg, i|
-  Facter.add("lvm_vg_#{i}") { setcode { vg } }
+  Facter.add("lvm_vg_#{i}") do
+    setcode { vg }
+  end
   Facter.add("lvm_vg_#{vg}_pvs") do
     setcode do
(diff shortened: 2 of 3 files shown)


vchepkov:
According to https://www.puppet.com/docs/puppet/7/fact_overview.html, there should not be code outside setcode

```
Note: Set all code inside the sections outlined above ⁠— there must not be any code outside setcodeand confine blocks other than an optional has_weight statement in a custom fact.
```

rtib:
That's true and one more cause why these legacy facts should be deprecated, as pointed out in Slack discussion by @cyberious
> Just note that when puppetlabs would be strict the lvm_vg_*_pvs, lvm_vg_* and lvm_pv_* facts should be marked as legacy facts and which are per default disabled with puppet 8. So maybe inventing time to use the new structured facts would be better in the long run.

They just need to work as long there is no alternative source for the particular information they provide and some time to migrate.

lib/facter/lvm_support.rb
```diff
@@ -17,47 +17,48 @@
...
   confine lvm_support: true
 
-  vgs = Facter::Core::Execution.execute('vgs -o name --noheadings 2>/dev/null', timeout: 30)
-
-  if vgs.nil?
-    setcode { 0 }
-  else
-    vg_list = vgs.split
-    setcode { vg_list.length }
+  if Facter.value(:lvm_support)
```
bastelfreak, line 20: this if should not be required, the `lvm_vgs` fact is already confineed on `lvm_support` (see line 18).

lib/facter/lvm_support.rb
```diff
@@ -17,47 +17,48 @@
...
 Facter.add('lvm_pvs') do
   confine lvm_support: true
 
-  pvs = Facter::Core::Execution.execute('pvs -o name --noheadings 2>/dev/null', timeout: 30)
-  if pvs.nil?
-    setcode { 0 }
-  else
-    pv_list = pvs.split
-    setcode { pv_list.length }
+  if Facter.value(:lvm_support)
```
bastelfreak, line 50: same here

lib/facter/lvm_support.rb
rtib: indeed, if it would be within the setcode block; I'm currently busy with getting rid off these nasty pieces here and to find a solution which works and passes spec tests

rtib:
The thing is as follows:

TL;DR
Those facts with programmatically generated names, i.e. `lvm_vg_<n>` and `lvm_vg_<vg_name>_pvs` rely on code outside any setcode blocks, but require the results of commands executed before. It is not possible to set global variables within a setcode block, that's why those commands need to be executed outside, where they are not guarded by confine statements. This obviously violates the requirement to have all code of this kind within setcode or confine blocks. They could, however, be accessed as fact values, but as they are needed to produce the names of the above facts, that would violate the requirement to have such code within setcode blocks only. Current RSpec implementation isn't able to deal with code outside setcode and confine statements, thus these features cannot be tested within spec. 

To me, all that sums up as the code as it is works for now, but cannot be tested thoroughly, and might break in near future. Any code, that can be tested and is future prove, won't be able to produce fact names programmatically - dispute me if I'm wrong.

I'd propose the following solution:
* leave the code within `lvm_support.rb` as it is in this PR, manually tested
* drop the failing spec tests
* deprecate the legacy facts having programmatically generated names now, and remove them as soon as possible
* implement replacement using a modern approach, i.e. structured/aggregated facts reporting the relations between pvs, vgs and lvs

lib/facter/lvm_support.rb
rtib: Please have a look at https://github.com/puppetlabs/puppetlabs-lvm/pull/308#issuecomment-1633927597

rtib:
That would work in the wild, though, it doesn't solve any problems we are facing here:

1. violating the format as described here: https://www.puppet.com/docs/puppet/7/fact_overview.html#writing_facts_simple_resolutions-how-to-format-facts
2. cannot write proper spec tests for

Moreover, `lvm_pv_<n>` can be replaced by using `physical_volumes`. The main focus is on the contents of `lvm_vg_<vg_name>_pvs`, which doesn't have an alternative source yet.

lib/facter/lvm_support.rb
```diff
@@ -17,47 +17,48 @@
...
+    vgs = Facter::Core::Execution.execute('vgs -o name --noheadings 2>/dev/null', timeout: 30)
+    vg_list = vgs.split unless vgs.nil?
   end
+
+  setcode { vg_list.length }
 end
 
-# lvm_vg_[0-9]+
-#   VG name by index
+# # lvm_vg_[0-9]+
```
malikparvez, line 28: any specific reason for double hashes?

rtib: No. It's the result of an IDE accident.

elfranne:
Tested on Windows 2019 too and works fine.

vchepkov:
I defer to experts in Puppet's internals. As I stated earlier, v2.0.3 and updated code doesn't conform to the documentation, code in v2.0.2 did.

README.md
```diff
@@ -346,6 +346,13 @@ Logical volume size can be extended, but not reduced -- this is for
 safety, as manual intervention is probably required for data
 migration, etc.
 
+## Deprecation Notice
+
+Some facts reported by this module are being deprecated in favor of upcomming structured facts.  The following facts are being deprecated:
```
Ramesh7, line 351: The below facts are the custom flat facts, so not sure if these are gonna deprecate as well?
AFAIK the custom facts (irrespective of flat or structured facts) these will not be deprecated. Only the [legacy facts](https://www.puppet.com/docs/puppet/7/core_facts.html#legacy-facts) which will get deprecated.

rtib: Facter code generating fact names dynamically on the fly is problematic: violates coding rules, hard to maintain, even harder to test. That's why we should get rid of them.

crazymind1337:
Can we please have this released soon?

Ramesh7 (approved):
we are moving with this change for now but for long term we need to review these facts with manifest use-case and update accordingly.

rajat-puppet:
These changes have been released as part of `2.1.0` release version

rtib:
Thank You.

Issue #298, opened 2023-06-14
Facts lvm_vg_* unavailable


rtib:
## Describe the Bug
Facts `lvm_vg_*` and `lvm_vg_*_pvs` aren't resolved since v2.0.0 release.

## Expected Behavior
As those facts where not declared deprecated neither removed from the code, they are expected to be available.

## Steps to Reproduce
1. Run `sudo puppet facts | grep lvm_vg` on a node having LVM volume groups configured with module version 2.0.x and earlier.

## Environment
 - Puppet agent version 7.24.0 (others may also affected)
 - Platform Debian 11 (other may also affected)

## Additional Context
With commit https://github.com/puppetlabs/puppetlabs-lvm/commit/8e1e8eb1718f31a75d9edf79b123c74d61364e55 the setcode blocks of the affected facts [here](https://github.com/puppetlabs/puppetlabs-lvm/blob/v2.0.0/lib/facter/lvm_support.rb#L30..L32) have been moved into the `setcode` block above. I assume, such an encapsulation of `setcode` blocks is not supported with Facter.

cruelsmith:
As @rtib also pointed out via slack the newer structured facts `logical_volumes`, `physical_volumes` and `volume_groups` are missing its relations to each other.
That should be also fixed by adding `pv_name`, `vg_name`, `lv_full_name` to the structured facts so it allows cross referencing.

PR #309, opened 2023-07-14
extend structured facts to provide relations between lvs, vgs and pvs

rtib:
## Summary
This is extending the fact `logical_volumes` with the name of the volume group they are residing on. An additional fact `volume_group_map` is providing a map of volume groups and their backing physical volumes.

## Additional Context
As discussed on Slack, in https://github.com/puppetlabs/puppetlabs-lvm/issues/298#issuecomment-1620164884 and boiled down in https://github.com/puppetlabs/puppetlabs-lvm/pull/308#issuecomment-1633927597, facts having programmatically generated name should be retired. These changes are necessary to close provide all information the deprecated facts were providing.

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/lib/facter/logical_volumes.rb
+++ b/lib/facter/logical_volumes.rb
@@ -26,9 +26,10 @@ Facter.add(:logical_volumes) do
       'lv_active',
       'lv_size',
       'lv_permissions',
+      'vg_name',
     ]
 
     output = Facter::Core::Execution.exec("lvs -o #{columns.join(',')}  --noheading --nosuffix")
-    Puppet_X::LVM::Output.parse('lv_name', columns, output)
+    Puppet_X::LVM::Output.parse('lv_name', columns, output, 'lv_')
   end
 end
--- a/lib/facter/volume_group_map.rb
+++ b/lib/facter/volume_group_map.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 Facter.add(:volume_group_map) do
-  # Fact should be confined to only linux servers that have the lvs command
+  # Fact should be confined to only linux servers that have the vgs command
   confine do
     Facter.value('kernel') == 'Linux' &&
       Facter::Core::Execution.which('vgs')
--- a/lib/facter/volume_group_map.rb
+++ b/lib/facter/volume_group_map.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+Facter.add(:volume_group_map) do
+  # Fact should be confined to only linux servers that have the lvs command
+  confine do
+    Facter.value('kernel') == 'Linux' &&
+      Facter::Core::Execution.which('vgs')
+  end
+
+  setcode do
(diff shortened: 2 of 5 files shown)


rtib:
The failed acceptance test broke down outside the test examples, might need to be kicked again.

cruelsmith:
First of all thanks for your work. :+1: 

In my eyes the goal should be to prevent any additional fact to add the relation between the `vg` and `pv`. It should be as simple as you provided it for the `lv`.
One way would be to rework the `Puppet_X::LVM::Output.parse` to something like this:
```
def self.parse(key, columns, data, prefix2remove)
  data.split("\n").map { |line|
    line.gsub(%r{\s+}, ' ').strip.split
  }.map { |line|
    columns.zip(line).to_h
  }.group_by { |h|
    h[key]
  }.transform_values { |v|
    tmp = {}
    columns.each { |column|
      tmp[column] = []
      v.each { |key,value|
        tmp[column] << key[column]
      }
      tmp[column].uniq!
      tmp[column] = tmp[column][0] if tmp[column].length == 1
    }
    tmp.transform_keys { |key|
      remove_prefix(key, prefix2remove)
    }
  }
end
```
After that the `pv_name` can be added to the argument list of the `volume_groups` facts.
Which then results in 
```
pv_name => [
      "/dev/sdb",
      "/dev/sdc",
      "/dev/sdd",
      "/dev/sde",
      "/dev/sdf"
    ],
```
or
```
pv_name => "/dev/sda",
```
to be added to the entries of the `volume_groups` fact.
----
Please note that this is only a PoC code, which means that it can probably still be optimized or that there still exist unfixed edge cases.

rtib:
Hi, @cruelsmith 

thank you for your suggestions. First, I also though that it would be simple enough to add an additional field to volume_groups revealing it's physical volumes. Then I stumbled over the strange output of `vgs` when extending it with `pv_name`, which actually multiply the rows having multiple pvs. Grouping that sounds straightforward, though, rise a couple of questions and concerns: which field to group by on? (uuid, vg_name?) What if there is a clustered or shared volume group? I don't even have the environment to test all those possibilities. 

Remarkably enough, there is no man page listing `pv_name` as valid field for reporting volume groups!

Such a solution would make lots of assumptions on the underlying API including undocumented features, requiring them to be stable among many versions, including future too. If something change, the whole thing might break.

While the same code which is generating `volume_group_map` might be added to `volume_groups.rb` to add an additional field, that would break the nice simplicity of that fact.

That's why I  was withdrawing from this approach and implemented a less attractive additional fact.

jst-cyr:
@rtib  : My apologies that nobody has responded on this for so much time. I'm going through our pull requests across modules and came across this request here. Is this still an issue for you?

rtib:
It has never been a personal issue to me. However, I've moved away from Puppet for now. My support goes either to open source software or to those who pay for it.

PR #330, opened 2024-02-09
pdk update to version 3.0.1 template-ref heads/main-0-g23c6fff

kenyon:


Gemfile
```diff
@@ -22,16 +22,15 @@ group :development do
...
   gem "rspec-puppet-facts", '~> 2.0',            require: false
-  gem "codecov", '~> 0.2',                       require: false
   gem "dependency_checker", '~> 1.0.0',          require: false
   gem "parallel_tests", '= 3.12.1',              require: false
   gem "pry", '~> 0.10',                          require: false
-  gem "simplecov-console", '~> 0.5',             require: false
+  gem "simplecov-console", '~> 0.9',             require: false
   gem "puppet-debugger", '~> 1.0',               require: false
-  gem "rubocop", '= 1.48.1',                     require: false
+  gem "rubocop", '~> 1.50.0',                    require: false
```
bastelfreak, line 33: upgrade rubocop but downgrade the rubytargetversion? Is that intentional?

kenyon: Not sure why `TargetRubyVersion` went to 2.7 in https://github.com/puppetlabs/puppetlabs-lvm/commit/f25a881d09a2c05244ad7cc5b90ded952a194e87#diff-4f894049af3375c2bd4e608f546f8d4a0eed95464efcdea850993200db9fef5cR8

It went from 2.5 to 2.6 in https://github.com/puppetlabs/pdk-templates/commit/79e331ededca639a9bad8f85523ae0deaa7884a6#diff-d5dfc512a1504644039a27b4ecabf1845748dbf0d47e21e2905a9ab5a45526aa which was in version 3.0.0 and 3.0.1 of pdk-templates.

PR #343, opened 2024-07-09
CAT-1912 : Adding new getters and setters for stripes & not allowing change of stripes for an existing lvm

malikparvez:
## Summary
Adding new getters and setters for stripes.

## Additional Context
Add any additional context about the problem here. 
- [ ] Root cause and the steps to reproduce. (If applicable)
- [ ] Thought process behind the implementation.

## Related Issues (if any)
Mention any related issues or pull requests.

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/lib/puppet/provider/logical_volume/lvm.rb
+++ b/lib/puppet/provider/logical_volume/lvm.rb
@@ -221,6 +221,22 @@ Puppet::Type.type(:logical_volume).provide :lvm do
     end
   end
 
+  def stripes
+    # Run the lvs command with the -o option to get only the stripes count
+    raw = (lvs '-o', 'stripes', '--noheadings', path)
+
+    output_array = raw.strip
+    output_array
+  end
+
+  def stripes=(new_stripes_count)
+    current_stripes = stripes.to_i
+
+    # Changing stripes is not supported for existing logical volumes
+    return unless new_stripes_count.to_i != current_stripes
+    raise(Puppet::Error, "Changing stripes from #{current_stripes} to #{new_stripes_count} is not supported for existing logical volumes")
+  end
+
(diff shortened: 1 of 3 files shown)


Ramesh7 (approved):

lib/puppet/provider/logical_volume/lvm.rb
```diff
@@ -221,6 +221,22 @@ def size=(new_size)
...
 
+  def stripes
+    # Run the lvs command with the -o option to get only the stripes count
+    raw = (lvs '-o', 'stripes', '--noheadings', path)
+
+    output_array = raw.strip
+    output_array
+  end
+
+  def stripes=(new_stripes_count)
```
bastelfreak, line 232: what's the purpose of this setter? Is that used anywhere?

malikparvez: Hey @bastelfreak , I was trying to not allow the user to change the stripes of an existing lvm so added setter to raise an error. Referred the implementation the way mirror & mirror= getter & setter handled. Please let me know if there is better way to do this happy to comply...

bastelfreak: ah okay. that was my impression as well. But than why did you change it to a gettter and setter at all?

malikparvez: I could only think of this way to handle it, can you suggest how to do this without getter and setter?

bastelfreak (approved):

lib/puppet/provider/logical_volume/lvm.rb
malikparvez: Discussed in DM, hence resolving it...

PR #346, opened 2024-07-11
MAINT : Removing support for centos 7,8 & debian 10

malikparvez:
## Summary
Removing support for centos 7,8 & debian 10.

## Additional Context
Add any additional context about the problem here. 
- [ ] Root cause and the steps to reproduce. (If applicable)
- [ ] Thought process behind the implementation.

## Related Issues (if any)
Mention any related issues or pull requests.

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -23,13 +23,6 @@
       ]
     },
     {
-      "operatingsystem": "CentOS",
-      "operatingsystemrelease": [
-        "7",
-        "8"
-      ]
-    },
-    {
       "operatingsystem": "Scientific",
       "operatingsystemrelease": [
         "7"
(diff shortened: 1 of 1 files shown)


metadata.json
```diff
@@ -22,13 +22,6 @@
         "9"
       ]
     },
-    {
-      "operatingsystem": "CentOS",
```
bastelfreak, line 26: did you check if there's also CentOS/Debian specific code that should be deleted?

malikparvez: yes, I did check could not find any code related to above os

bastelfreak (approved):
I'm not sure if maint is the correct label here, but I'm fine with the code changes itself.

malikparvez:
> I'm not sure if maint is the correct label here, but I'm fine with the code changes itself.

Added backwards-incompatible, as cannot call it a feature or a bug either...

PR #380, opened 2026-04-30
(MODULES-11756) Wait for udev to settle after lvcreate

imaqsood:
## Summary

Fixes [MODULES-11756](https://perforce.atlassian.net/browse/MODULES-11756).

When a manifest declares a `logical_volume` followed immediately by a `filesystem` resource targeting the same device, `mkfs` can fail with **`device or resource busy`**. The cause: `lvcreate` returns before udev has finished processing device-add events for the new LV — udev rules (symlink creation, blkid scans, multipath, lvmetad notifications) still hold the device open briefly, racing with `mkfs`'s exclusive open. Re-running the agent succeeds because the device has settled by then.

The fix is to call `udevadm settle --timeout=30` after a successful `lvcreate`, exactly as LVM2's own user-space tools do for the same reason. This is the deterministic mechanism — not a fixed sleep, not a tunable, no per-environment guessing. The 30s bound prevents a stuck udev daemon from hanging the agent.

This is a **patch-level, behaviour-preserving** fix. No new public parameter, no API change, no manifest changes required. Existing manifests run identically — the only observable difference is that the race no longer occurs.

### Behaviour

| Condition | Behaviour |
|---|---|
| `udevadm` present, settle succeeds | Silent success |
| `udevadm` not on PATH | `debug` log, no settle, `create` still succeeds |
| `udevadm settle` exits non-zero (e.g. timeout) | `warning` log, no raise — a flaky settle should not turn a previously-working resource into a failing one |
| `lvcreate` fails | Settle is not called |
| LV already exists / `destroy` | Settle is not called |

### AIX

**AIX behaviour is unchanged.** AIX does not use udev; `Puppet::Util.which('udevadm')` returns `nil`, the debug-log branch fires, and `create` is a no-op for the new code path.

## Changes

- `lib/puppet/provider/logical_volume/lvm.rb` — register `udevadm` as an optional command; call `settle_udev` after a successful `lvcreate`; cache `udevadm` availability per provider instance.
- `spec/unit/puppet/provider/logical_volume/lvm_spec.rb` — new specs for the four behaviours above plus a destroy-path assertion.
- `CHANGELOG.md` — Fixed entry under v3.0.2.
- `metadata.json` — patch bump 3.0.1 → 3.0.2.

## Test plan

- [x] Unit tests: `bundle exec rake spec` — 171 examples, 0 failures.
- [x] `bundle exec rake validate` — clean.
- [x] `bundle exec rake rubocop` — 51 files inspected, 0 offenses.
- [ ] **Acceptance test on a litmus runner with a real LVM stack** is desirable as a follow-up but is out of scope for this PR (no regression risk: the failure mode being fixed is itself an acceptance-only race).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


imaqsood:
## Code review (Claude)

**Verdict:** approve-with-comments. Tests pass (203/0), rubocop / puppet-lint clean, spec adherence is tight.

### Must-fix
None.

### Should-fix

1. **Cache is per-resource, not per-run.** `@udevadm_available` lives on the provider *instance*, but Puppet creates a fresh instance per resource per run — so we re-`which` for every LV. The spec literally says "cache for the lifetime of the provider instance," so this conforms, but a class-level (`@@`) cache would be a one-liner improvement.
2. **`optional_commands udevadm` is redundant** with the explicit `Puppet::Util.which` guard. Both belt-and-braces; harmless but slightly noisy. Pick one.
3. **`lvcreate fails → settle not called` test is a bit weak.** It would pass even if someone reordered `settle_udev` above `lvcreate`, because `lvcreate` raises before either path is reached. The `which` stub closes most of the hole, but a tighter test could expect `:lvcreate` and `:udevadm` in a `sequence` and assert the sequence fails at lvcreate.

### Nits

- Warning includes `@resource[:name]` as the spec mandates ✓
- Debug string matches spec verbatim ✓
- `defined?(@udevadm_available)` correctly handles cached `nil` (a plain `@x ||=` would re-check forever when udevadm is absent) ✓
- CHANGELOG / metadata format matches v4.0.0 entries; patch bump 4.0.0 → 4.0.1 ✓
- No files touched outside spec §6 in-scope list ✓

### Specifically verified

- `settle_udev` runs only after a successful `lvcreate` in `create` — not in `destroy`, not in property-sync, not when `exists?` is true.
- `Puppet::ExecutionFailure` is the right rescue class (consistent with lines 137 / 210 / 310 of the same file).
- `--timeout=30` (not configurable, not `30s`) per spec §3 / §7.1.
- Destroy spec `unstub`s `:settle_udev` so it actually exercises the live path.
- The global `before(:each)` stub of `:settle_udev` cleanly insulates pre-existing create-path tests; the new context unstubs it.

**Recommendation:** ship as-is, or address should-fix #1 (class-level cache) in the same PR.

🤖 Review by [Claude Code](https://claude.com/claude-code)

amitkarsale (approved):

imaqsood:
Follow-up commit `044f91e` addresses two of the three should-fixes from the code review above:

1. ✅ **Cache moved to class level** — `@udevadm_available` now lives on the provider class, so the binary lookup runs once per process rather than once per resource.
2. ✅ **Single source of truth for `udevadm` path** — replaced `Puppet::Util.which('udevadm')` with `command(:udevadm)` (the resolver registered by `optional_commands` above). No more two-ways-of-asking-the-same-question.
3. ✅ **Tightened `lvcreate fails → settle not called` test** with a Mocha `sequence` so a regression that reordered `settle_udev` before `lvcreate(*args)` would also be caught.

No behaviour change. Tests updated to stub `provider_class.udevadm_available?` (matching the new source of truth) instead of `Puppet::Util.which`. 203 examples, 0 failures locally.

PR #384, opened 2026-06-09
(MODULES-11840) Allow puppetlabs/stdlib 10.x

imaqsood:
Part of MODULES-11840. Metadata-only change: stdlib dependency cap `< 10.0.0` -> `< 11.0.0` so this module installs alongside stdlib 10.x. No code change.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -10,7 +10,7 @@
   "dependencies": [
     {
       "name": "puppetlabs/stdlib",
-      "version_requirement": ">= 4.13.1 < 10.0.0"
+      "version_requirement": ">= 4.13.1 < 11.0.0"
     }
   ],
   "operatingsystem_support": [


bastelfreak (dismissed):
Can we please first discuss what's going to be in stdlib 10?

imaqsood:
> Can we please first discuss what's going to be in stdlib 10?

Thanks for flagging. We're planning to proceed with the stdlib 10 release, which is why these modules need their dependency cap lifted from < 10.0.0 to < 11.0.0 so they remain installable alongside it. This PR is a metadata-only change to enable that.

For more context on what's going into stdlib 10 and the release direction, please connect with Jason St-Cyr.

jst-cyr:
@bastelfreak : The release will contain the changes currently listed in tag 10.0.0 of the stdlib library, though we will be releasing it as 10.0.1 on the Forge after pulling back the 10.0.0 release. I will be publishing an article to announce the upcoming release more widely to give some notice, but you can see the specific commits in the stdlib repository.

The puppetlabs modules that depend on stdlib will be updated to allow for them to use the new version when it is released.

bastelfreak:
And why does perforce want to do a major release with a single breaking change that's avoidable, and why doesn't perforce remove the 24 deprecated functions?

jst-cyr:
> And why does perforce want to do a major release with a single breaking change that's avoidable, and why doesn't perforce remove the 24 deprecated functions?

@bastelfreak : Hopefully the article I posted today on puppet.com and dev.to will help with explaining the major release: https://www.puppet.com/blog/puppetlabs-stdlib-10

In terms of the deprecated functions, that seems like a discussion to have outside the context of this particular PR, but it does sound like something that should also be addressed in a major release.

mehul-jain1 (approved):

jst-cyr:
@imaqsood and @mehul-jain1 : Are we able to merge this one up?

PR #390, opened 2026-07-04
Add RHEL10 support  (#372)

jst-cyr:
## Summary
The module should manage RHEL10 correctly, and remove edit the existing swap entry instead of adding a new one.
This pull request adds RHEL 10 into metadata to trigger acceptance testing and adds specific acceptance tests to validate that swap resize functions.
TBD: If swap resize is still failing, this pull request needs to be updated to also correct the swap resize issue.

## Additional Context

From #372:
> Since RHEL10 is using partition UUID for everything (including a kernel cmdline setting for hibernate) the module does not work as expected. After resizing the swap I get:
> 
> ```
> UUID=40a82883-0ad6-45de-856a-27f845f144ea       none    swap    defaults        0       0
> /dev/vg0/swap   swap_/dev/vg0/swap      swap    defaults        0       0
> 
> ```
> (and I need to edit the kernel cmdline to change/remove resume=UUID=XXXX as documented in a Red Hat KB, I personnally removed the line in a kickstart postinstall script to avoid this issue forever).

## Related Issues (if any)
Addresses #372 issues for RHEL 10. 

## Checklist
- [ ] 🟢 Spec tests.
- [X] 🟢 Acceptance tests.

### DIFF
--- a/lib/puppet/provider/logical_volume/lvm.rb
+++ b/lib/puppet/provider/logical_volume/lvm.rb
@@ -205,7 +205,10 @@ Puppet::Type.type(:logical_volume).provide :lvm do
             mount_point = lsblk('-o', 'MOUNTPOINT', '-nr', path).chomp
             xfs_growfs(mount_point) || raise("Cannot resize filesystem to size #{new_size} because xfs_growfs failed.")
           elsif %r{\bTYPE="(swap)"}.match?(blkid_type)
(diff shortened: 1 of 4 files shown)


jst-cyr:
This is still a draft. Currently adding a new swap acceptance test that only runs on RHEL 10 to test for the issue raised in #372.

jst-cyr:
I have added a test that represents the issue found in #372 with swapping and fix to preserve the UUID. Moved this from draft to ready for review so that the team can run the acceptance tests when available.

